PR #38, opened 2020-11-03
Command output capture options

mfranciszkiewicz:
TODO:
- address 2 `FIXME`s in code (now defaults to output streaming)
- wait for `ya-ts-client` 0.2.x


prekucki:
TODO: Change it to spec to the new activity model.

PR #39, opened 2020-11-04
Implement summary logger

### LINKED ISSUE #27: yajsapi catching up with yapapi: Logs as event emmiter

mdtanrikulu:
Resolves #27


yajsapi/utils/log.ts
```diff
@@ -1,37 +1,303 @@
...
+      console.log("error", error);
+      logger.error("SummaryLogger entered invalid state");
+      this.error_occurred = true;
+    }
+  }
+
+  _handle(event: events.YaEvent) {
+    const eventName = event.constructor.name;
+    if (eventName === events.ComputationStarted.name) this._reset();
+    if (eventName === events.SubscriptionCreated.name)
```
azawlocki, line 166: Why not use `switch(eventName)` here, instead of a sequence of `if`s?

azawlocki (approved):
I've left some comments but overall it looks good!

examples/blender/blender.ts
```diff
@@ -1,7 +1,8 @@
 import path from "path";
 import dayjs from "dayjs";
 import duration from "dayjs/plugin/duration";
-import { Engine, Task, utils, vm, WorkContext } from "yajsapi";
+import { Engine, Task, utils, vm, WorkContext } from "../../yajsapi";
+import { logSummary } from "../../yajsapi/utils/log";
```
filipgolem, line 5: `../../` should be removed before merging.

yajsapi/utils/log.ts
```diff
@@ -1,37 +1,303 @@
...
+    } else if (eventName === events.AgreementConfirmed.name) {
+      logger.info(
+        `Agreement confirmed by provider '${
+          this.agreement_provider_name[event["agr_id"]]
+        }'`
+      );
+      this.confirmed_agreements.add(event["agr_id"]);
+    } else if (eventName === events.TaskStarted.name)
+      this.task_data[event["task_id"]] = event["task_data"];
+    else if (eventName === events.ScriptSent.name) {
```
filipgolem, line 213:
```suggestion
    } else if (eventName === events.TaskStarted.name) {
      this.task_data[event["task_id"]] = event["task_data"];
    } else if (eventName === events.ScriptSent.name) {
```

filipgolem (approved):

yajsapi/utils/log.ts
```diff
@@ -1,37 +1,303 @@
...
-}
+  info: "blue",
+  debug: "magenta",
+  warn: "yellow",
+  error: "red",
+};
 winston.addColors(logColors);
 const logger = winston.createLogger(options);
 
+class SummaryLogger {
```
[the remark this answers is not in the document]
mdtanrikulu: I am in favor of having different options to the same instance than different instances for different methods here.
But we can create child instance based on default one to override some options like meta for each method  i.e. `defaultMeta: { method: "METHOD_NAME" },`
https://github.com/winstonjs/winston#creating-child-loggers

yajsapi/utils/log.ts
mdtanrikulu: That's right, it can be used here. but I was actually thinking about "lookup table". much performant and may decrease the code line and keep messages and functions organized in an object. if/else was just fastest option without thinking much :)

PR #43, opened 2020-11-06
Implement multi payment platform

mdtanrikulu:


yajsapi/rest/payment.ts
```diff
@@ -196,12 +210,26 @@ export class Payment {
...
+  }
+
+  async decorate_demand(ids: string[]): Promise<yap.MarketDecoration> {
+    const { data: _decorate_demand } = await this._api.decorateDemand(ids);
+    return _decorate_demand; 
+  }
+
   async *invoices(): AsyncGenerator<Invoice> {
     let { data: result } = await this._api.getReceivedInvoices();
     for (let invoice_obj of result) {
```
filipgolem, line 235: In `accounts()` the loop looks like this:
`for await ( let account_obj of _accounts ) {`
but in `invoices()` it looks like this:
`for (let invoice_obj of result) {`

mdtanrikulu: nice catch, thanks! it should be without `await` in `accounts`.

filipgolem (approved):

PR #46, opened 2020-11-16
Make yajsapi depend on "^0.2.0", not "file:../"

filipgolem:


examples/package.json
```diff
@@ -15,7 +15,7 @@
         "bluebird": "^3.5.0",
         "dayjs": "^1.8.31",
         "ts-node": "^9.0.0",
-        "yajsapi": "file:../"
+        "yajsapi": "^0.2.0"
```
mdtanrikulu, line 18: afaik it should be with `-alpha.0` suffix.

PR #67, opened 2020-11-24
More like yapapi

filipgolem:


yajsapi/runner/index.ts
```diff
@@ -695,9 +697,17 @@ export class Engine {
...
+        let done_workers = new Set([...workers].filter((worker) => !worker.isPending()));
+
         workers = new Set([...workers].filter((worker) => worker.isPending()));
         services = services.filter((service) => service.isPending());
 
+        for (let worker_task of [...done_workers]) {
+            if (worker_task.isPending()) {
+                await worker_task;
+            }
+        }
```
azawlocki, line 709: This does nothing: `done_workers` is the set of workers for which `worker.isPending()` is false, so the condition in `if` is never satisfied (assumming a worker cannot change from `! isPending()` to `isPending()`).

yajsapi/runner/index.ts
```diff
@@ -719,21 +728,27 @@ export class Engine {
           for (let worker_task of [...workers]) {
             worker_task.cancel();
           }
-          // await asyncio.wait(workers, timeout=15, return_when=asyncio.ALL_COMPLETED)
+          await bluebird.Promise.any([
+            bluebird.Promise.all([...workers]),
+            promise_timeout(10)
+          ]);
```
azawlocki, line 734: I don't know `bluebird.Promise`, but this looks suspicious: shouldn't it be `bluebird.Promise.all(...)`?

yajsapi/runner/index.ts
```diff
@@ -719,21 +728,27 @@ export class Engine {
...
       await bluebird.Promise.any([
-        Promise.all([process_invoices_job]),
-        promise_timeout(15),
+        bluebird.Promise.all([
+          find_offers_task,
+          process_invoices_job,
+        ]),
+        promise_timeout(10)
       ]);
+      if (agreements_to_pay.size == 0) {
```
azawlocki, line 746: In `yapapi` we wait for `process_invoices_job` only if `agreements_to_pay` is **not empty**.

azawlocki (changes requested):

yajsapi/runner/index.ts
filipgolem: Good catch!

filipgolem: Changed to > 0

yajsapi/runner/index.ts
filipgolem: Removed.

azawlocki (approved):

PR #70, opened 2020-11-25
Fix tasks invoices handling

filipgolem:


yajsapi/utils/log.ts
```diff
@@ -135,15 +137,22 @@ class SummaryLogger {
...
+        const tasks = this.provider_tasks[name];
+        const cost = this.provider_cost[name] || "0 (no invoices?)";
+        return {
+          'Provider Name': name,
+          'Tasks Computed': tasks ? tasks.length : 0,
+          'Total Cost': cost,
+        };
+      }
+    );
+    const total_cost = Object.values(this.provider_cost).reduce((item, acc) => (acc + item), 0);
```
mdtanrikulu, line 152: One little correction, seems we missed, `accumulator` for the reduce should be on the left side. Current one works cuz anyway we return result of sum, but conceptually it would be better to keep in the right position.
```js
    const total_cost = Object.values(this.provider_cost).reduce((acc, item) => (acc + item), 0);
```

mdtanrikulu (approved):

PR #76, opened 2020-11-30
Sign new agreements only if there are unassigned tasks

### LINKED ISSUE #75: yajsapi is sometimes signing agreements after the computation is finished.
Fixed by #76

filipgolem:
Fixes #75

### DIFF
--- a/yajsapi/runner/index.ts
+++ b/yajsapi/runner/index.ts
@@ -609,7 +609,7 @@ export class Engine {
         if (
           Object.keys(offer_buffer).length > 0 &&
           workers.size < self._conf.max_workers &&
-          work_queue.has_unassigned_tasks()
+          work_queue.has_unassigned_items()
         ) {
           let _offer_list = Object.entries(offer_buffer);
           let _sample =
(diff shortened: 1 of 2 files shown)


yajsapi/runner/index.ts
```diff
@@ -608,7 +608,8 @@ export class Engine {
         await sleep(2);
         if (
           Object.keys(offer_buffer).length > 0 &&
-          workers.size < self._conf.max_workers
+          workers.size < self._conf.max_workers &&
+          work_queue.has_unassigned_tasks()
```
azawlocki, line 612: `work_queue.has_unassigned_items()` would be better IMO, since this class uses `items` in several field names (they're not part of the public API but still)

azawlocki (approved):
I have only one comment regarding the name of the new method.

azawlocki (commented):
We should discuss if we want to add these changes. See the discussion in the description of a similar PR for `yapapi`: https://github.com/golemfactory/yapapi/pull/162

azawlocki (commented):
We should discuss if we want to add these changes. See the discussion in the description of a similar PR for yapapi: golemfactory/yapapi#162

yajsapi/runner/index.ts
filipgolem: Good idea!

PR #81, opened 2020-12-04
Add  typedoc, configure for html & markdown

### LINKED ISSUE #30: Add yajsapi docstrings (tsdoc)

mdtanrikulu:
Resolves #30.


yajsapi/props/base.ts
```diff
@@ -111,6 +117,18 @@ export class Model {
...
 
+  /**
+   * Initialize the model from an object representation.
+   * 
+   * @description When provided with an object of properties, it will find the matching keys
+        within it and fill the model fields with the values from the object.
+        
+        It ignores non-matching keys - i.e. doesn't require filtering of the properties'
+        object before the model is fed with the data. Thus, several models can be
+        initialized from the same object and all models will only load their own data.
```
azawlocki, line 128: Maybe add leading `*`'s?

yajsapi/props/builder.ts
```diff
@@ -1,6 +1,16 @@
 import dayjs from "dayjs";
 import { Market, Subscription } from "../rest/market";
 
+/**
+ * Builds an object of properties and constraints from high-level models.
+ * @example
+ * @description The object represents a Demand object, which is later matched by the new Golem's
+    market implementation against Offers coming from providers to find those providers
+    who can satisfy the requestor's demand.
+ *
```
azawlocki, line 10:
```suggestion
 * Builds an object of properties and constraints from high-level models.
 * 
 * @example  TODO
 *
 * @description The object represents a Demand object, which is later matched by the new Golem's
 *  market implementation against Offers coming from providers to find those providers
 *  who can satisfy the requestor's demand.
 * 
```

azawlocki (approved):
Left some minor comments on formatting

package.json
```diff
@@ -6,8 +6,10 @@
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "scripts": {
-    "clean": "rimraf node_modules/ dist/ || echo 'Project is already clean'",
     "build": "tsc",
+    "clean": "rimraf node_modules/ dist/ || echo 'Project is already clean'",
```
shadeofblue, line 10: @mdtanrikulu I'd say we should also have a (separate) command here to clean the `docs` directory -> especially since both `docs` and `docs:md` write to the same destination directory

yajsapi/executor/events.ts
```diff
@@ -3,16 +3,25 @@ import applyMixins from "../utils/applyMixins";
 
 type ExcInfo = [Error, Error, any?]; // TracebackType
 
+/**
+ * An abstract base class for types of events emitted by `Engine.map()`.
```
shadeofblue, line 7: I think that's `Executor.submit()`

yajsapi/props/base.ts
```diff
@@ -94,6 +94,12 @@ class InvalidPropertiesError extends Error {
       this.name = key;
     }
 }
+
+/**
+ * Base class from which all property models inherit.
+ * 
+ * @description Provides helper methods to load the property model data from an object and to get a mapping of all the keys available in the given model.
```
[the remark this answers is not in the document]
mdtanrikulu: there is "ignoreComments" option for eslint, but if distractive we can include it.

package.json
mdtanrikulu: Calling `rm -rf docs` or `yarn clean:docs` doesn't really make a difference in a sense of complexity. But if we want everyone in the same page, sure I can add that. Opinions?

yajsapi/executor/events.ts
mdtanrikulu: good catch, thanks!

package.json
shadeofblue: well, at least it eliminates any doubt

PR #89, opened 2020-12-09
Rename, refactor the codebase

### LINKED ISSUE #84: Renaming components in yajsapi
ensure names of classes and methods are consistent with what we have in yapapi

mdtanrikulu:
Resolves #84.


yajsapi/executor/index.ts
```diff
@@ -170,12 +73,22 @@ export class _BufferItem {
 type D = "D"; // Type var for task data
 type R = "R"; // Type var for task result
 
-export class Engine {
+export type ExectorOpts = {
```
azawlocki, line 76: why not `ExecutorOpts`?

azawlocki (approved):
I've read it all and it looks good as far as I can tell. Haven't run it though.

yajsapi/executor/index.ts
mdtanrikulu: oh typo :)  thanks for catching it.

mfranciszkiewicz (approved):

yajsapi/executor/strategy.ts
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  async score_offer(offer: OfferProposal): Promise<Number> {
+    const linear: ComLinear = new ComLinear().from_properties(offer.props());
+
+    if (linear.scheme.value != BillingScheme.PAYU) {
+      return SCORE_REJECTED;
+    }
+
+    if (linear.fixed_price > this.max_fixed) return SCORE_REJECTED;
```
filipgolem, line 54: This is not related to renaming, but it's risky to reject offers with an arbitrarily chosen `0.05` initial price. Is this value high or low?

filipgolem (approved):

yajsapi/executor/strategy.ts
mdtanrikulu: This seems important, I suggest create an issue for it.

PR #90, opened 2020-12-11
Fix long computations failing on providers

filipgolem:
- [x] handle 408 errors
- [x] improve logging
- [x] decrease getExecBatchResults timeout (fixes connection closed/ECONNRESET problem)
- [x] decrease sleep duration between API calls (fixes connection closed/ECONNRESET problem)
- [x] move `if (!any_new)` to the correct place

### DIFF
--- a/yajsapi/rest/activity.ts
+++ b/yajsapi/rest/activity.ts
@@ -137,11 +137,13 @@ class Activity {
   protected _state!: RequestorStateApi;
   protected _id!: string;
   protected _credentials?: object;
+  _finished: boolean;
 
   constructor(id: string, _api: RequestorControlApi, _state: RequestorStateApi) {
     this._id = id;
     this._api = _api;
     this._state = _state;
+    this._finished = false;
   }
 
   set id(x) {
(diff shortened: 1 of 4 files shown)


yajsapi/rest/activity.ts
```diff
@@ -344,8 +344,16 @@ class Batch implements AsyncIterable<Result> {
     let last_idx = 0;
     while (last_idx < this._size) {
       let any_new: boolean = false;
-      let exe_list = await this._activity.results(this._batch_id);
-      let results: yaa.ExeScriptCommandResult[] = exe_list;
+      let results: yaa.ExeScriptCommandResult[] = []
+      try {
+        results = await this._activity.results(this._batch_id);
+      } catch (error) {
+        if (error.response && error.response.status && error.response.status == 408) {
```
mdtanrikulu, line 351: Afair, it can be done with `?` optional chaining in a more readable way, but not necessarily important. can remain as it is.
`if( error?.response?.status == 408 )`

mdtanrikulu (approved):

yajsapi/rest/activity.ts
filipgolem: Available in Node.js 14, we are supporting Node.js 12.

PR #109, opened 2021-01-25
Add logging to linear market strategy class

filipgolem:


yajsapi/executor/strategy.ts
```diff
@@ -76,26 +76,38 @@ export class LeastExpensiveLinearPayuMS {
...
+      logger.debug(`Rejected offer ${offer.id()}: unsupported scheme '${linear.scheme.value}'`);
+      return SCORE_REJECTED;
+    }
 
-    const known_time_prices = [Counter.TIME, Counter.CPU];
+    const known_time_prices = new Set([Counter.TIME, Counter.CPU]);
 
-    for (const counter in Object.keys(linear.price_for)) {
-      if (!(counter in known_time_prices)) return SCORE_REJECTED;
+    for (const counter of Object.keys(linear.price_for)) {
```
mdtanrikulu, line 87: one more correction here, 
`for (const counter of Object.keys(linear.price_for))` same as  `for (const counter in linear.price_for)`
I would suggest having second one, instead of additional `Object.keys()` call

mdtanrikulu (approved):

PR #125, opened 2021-02-11
Command output logs & getExecBatchResults timeout fix

filipgolem:
No changes when `-d` is not specified.
When it's specified, all command output (stdout and stderr) is appended to the log file. If a given command fails on a provider, its output is additionally printed to the console.


rad9k (approved):
testes. seems OK.

PR #126, opened 2021-02-11
Market strategy that decreases score for providers that fail to confirm agreements

filipgolem:
See: https://github.com/golemfactory/yapapi/pull/246

### DIFF
--- a/yajsapi/executor/index.ts
+++ b/yajsapi/executor/index.ts
@@ -33,7 +33,13 @@ export const sgx = _sgx;
 export const vm = _vm;
 import { Task, TaskStatus } from "./task";
 import { Consumer, SmartQueue } from "./smartq";
-import { LeastExpensiveLinearPayuMS, MarketStrategy, SCORE_NEUTRAL } from "./strategy";
+import {
+  ComputationHistory,
+  DecreaseScoreForUnconfirmedAgreement,
+  LeastExpensiveLinearPayuMS,
+  MarketStrategy,
+  SCORE_NEUTRAL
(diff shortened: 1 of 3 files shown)


yajsapi/executor/strategy.ts
```diff
@@ -111,3 +115,32 @@ export class LeastExpensiveLinearPayuMS {
...
+
+  async decorate_demand(demand: DemandBuilder): Promise<void> {
+    /* Decorate `demand` using the base strategy. */
+    await this._base_strategy.decorate_demand(demand);
+  }
+
+  async score_offer(offer: OfferProposal, history?: ComputationHistory): Promise<Number> {
+    /* Score `offer` using the base strategy and apply penalty if needed.
+       If the offer issuer failed to approve the previous agreement (if any)
+       then the base score is multiplied by `self._factor`. */
```
azawlocki, line 138:
```suggestion
       then the base score is multiplied by `this._factor`. */
```

yajsapi/executor/index.ts
```diff
@@ -129,6 +135,7 @@ export class Executor {
   private _expires;
   private _budget_amount;
   private _budget_allocations: Allocation[];
+  private _rejecting_providers: Set<string>;
```
azawlocki, line 138: In the Python version the corresponding set is stored in `AgreementsPool` which is created for each `Executor.submit()` call. Here one set is used through possibly many calls to `Executor.submit()`.

I guess it's fine, maybe even better than in the Python version?

azawlocki (approved):

PR #138, opened 2021-02-24
Add computation price caps

filipgolem:
• Add computation price caps
• Fix HTTP 400 on `.reject()`
• Update default Executor timeout to 15 min (yapapi PR #255)
• Bug fix (`stdout` -> `stderr`)

### DIFF
--- a/yajsapi/executor/strategy.ts
+++ b/yajsapi/executor/strategy.ts
@@ -69,8 +69,10 @@ export class DummyMS extends MarketGeneral {
 
 export class LeastExpensiveLinearPayuMS {
   private _expected_time_secs: number;
-  constructor(expected_time_secs: number = 60) {
+  private _price_caps?: ComLinear;
+  constructor(expected_time_secs: number = 60, price_caps?: ComLinear) {
     this._expected_time_secs = expected_time_secs;
+    if (price_caps) this._price_caps = price_caps;
   }
 
   async decorate_demand(demand: DemandBuilder): Promise<void> {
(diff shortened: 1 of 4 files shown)


yajsapi/executor/strategy.ts
```diff
@@ -69,8 +69,10 @@ export class DummyMS extends MarketGeneral {
 
 export class LeastExpensiveLinearPayuMS {
   private _expected_time_secs: number;
-  constructor(expected_time_secs: number = 60) {
+  private _price_caps?: ComLinear;
+  constructor(expected_time_secs: number = 60, price_caps?: ComLinear) {
```
azawlocki, line 73: `fixed_price_cap` in `ComLinear` is declared as `fixed_price!: number;` so it's supposed to be not `null` or `undefined`.
In the Python original, it's `float`, so not `None` either. Thus I'm not sure if using `ComLinear` to represent user constraints, which are optional, is a good idea: what if the user does not want to put any constraints on the fixed price?

azawlocki: Perhaps it would be better to use two separate parameters: `max_fixed_price?: number` and `max_price_for?: Map<com.Counter, number>`?

filipgolem: Fixed as you suggested.

azawlocki (approved):
Good work with catching the bug in the call to `rejectProposalOffer`!

yajsapi/rest/market.ts
```diff
@@ -145,10 +145,16 @@ export class OfferProposal {
...
   async reject(_reason: string | null = null) {
-    await this._subscription._api.rejectProposalOffer(
-      this._subscription.id(),
-      this.id()
-    );
+    try {
+      await this._subscription._api.rejectProposalOffer(
+        this._subscription.id(),
+        this.id(),
+        { message: "Rejected" as Object }
```
filipgolem, line 152:
```suggestion
        { message: _reason as Object }
```

PR #140, opened 2021-03-01
Break from async generator

filipgolem:


yajsapi/storage/gftp.ts
```diff
@@ -80,10 +80,14 @@ class GftpDriver {
...
+      if (result === undefined) {
+        throw value;
+      }
       return result;
     } catch (error) {
-      logger.error(`gftp error. query: ${query} value: ${valueStr} error: ${JSON.stringify(error)}`);
-      throw Error(error);
+      const msg = `gftp error. query: ${query} value: ${valueStr} error: ${JSON.stringify(error)}`;
+      logger.error(msg);
+      throw Error(msg);
```
filipgolem, line 90:
```suggestion
      throw Error(error);
```

yajsapi/executor/index.ts
```diff
@@ -800,7 +833,13 @@ export class Executor implements ComputationHistory {
       }
       emit(new events.ComputationFinished());
     } catch (error) {
-      logger.error(`fail= ${error}`);
+      if (error instanceof AsyncGeneratorBreak) {
+        work_queue.close();
+        done_queue.close();
+        logger.info("User decided to stop getting new results.");
```
azawlocki, line 839: I'm not sure I like this message. Why mention "User" and imply any "decision"? The computation is stopped simply by a `break` statement in code.

azawlocki (approved):
Looks good! I've left one minor comment

yajsapi/executor/index.ts
```diff
@@ -759,7 +788,11 @@ export class Executor implements ComputationHistory {
     ];
     try {
       while (services.indexOf(wait_until_done) > -1 || !done_queue.empty()) {
-        if (cancellationToken.cancelled) { work_queue.close(); done_queue.close(); break; }
+        if (cancellationToken.cancelled) {
+          work_queue.close();
+          done_queue.close();
```
azawlocki, line 793: These two statements are repeated in `catch (error)` clause, it would be good to have them in one place, since both kinds of cancellation (caused by ctrl+c and by the `break` statement) should probably be handled in the same way.

yajsapi/executor/index.ts
filipgolem, line 839:
```suggestion
        logger.info("Break in the async for loop. Gracefully stopping all computations.");
```

PR #146, opened 2021-03-10
Add GitHub action that runs unit tests

filipgolem:


.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,31 @@
...
+      - b0.*
+  pull_request:
+    branches:
+      - master
+      - b0.*
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
```
shadeofblue, line 16: don't we _also_ want to run it on different platforms? windows and macos?

filipgolem: Fixed

shadeofblue (approved):

PR #155, opened 2021-03-30
Fix market timeouts

filipgolem:


yajsapi/executor/index.ts
```diff
@@ -465,17 +470,16 @@ export class Executor implements ComputationHistory {
...
-                reason: reason,
-              }));
             } catch (error) {
-              //suppress and log the error and continue;
               logger.log("debug", `Reject error: ${error}`);
             }
+            emit(new events.ProposalRejected({
+              prop_id: proposal.id(),
+              reason: reason,
+            }));
```
azawlocki, line 482: I'd prefer if `ProposalRejected` were emitted only when `proposal.reject()` succeeds. If it fails then we can emit `ProposalFailed`. This would be consistent with emitting `ProposalResponded` when `proposal.respond()` succeeds and `ProposalFailed` otherwise.

azawlocki (approved):
You're good to go!

PR #159, opened 2021-04-06
Fix path

filipgolem:
Support both `../../dist/` and `../../dist/yajsapi/` as build paths.


azawlocki (approved):
Please do not forget to squash

PR #172, opened 2021-04-12
Port agreement pool from yapapi

filipgolem:


yajsapi/executor/agreements_pool.ts
```diff
@@ -0,0 +1,192 @@
...
+      }
+    } catch (e) {
+      emit(
+        new events.ProposalFailed({
+          prop_id: offer.proposal.id(),
+          reason: e.toString(),
+        })
+      );
+      throw e;
+    }
```
azawlocki, line 143: I'd consider rewriting this:
```javascript
try {
   const agreement = await offer.proposal.create_agreement();
   try { 
      <code>
   } catch (error) {
      ... return;
   }
} catch (e) {
   ... throw e;
}
```
to:
```javascript
try {
   const agreement = await offer.proposal.create_agreement();
} catch (e) {
   ... throw e;
}
try {
   <code>
} catch (error) {
   ... return;
}
```
to avoid nesting `try ... catch` statements.

yajsapi/executor/index.ts
```diff
@@ -419,90 +412,105 @@ export class Executor implements ComputationHistory {
       logger.debug("Stopped processing debit notes.");
     }
 
-    async function find_offers(): Promise<void> {
-      let _subscription: Subscription;
+    async function find_offers_for_subscription(subscription: Subscription): Promise<void> {
```
azawlocki, line 415: Does this PR also contain changes not related to agreement pool? If so, it would be useful to list them in the PR description.

azawlocki (changes requested):
`yapapi` has two integration test cases that test features related to changes in this PR:
* [`test_multiactivity_agreement`](https://github.com/golemfactory/yapapi/blob/master/tests/goth/test_multiactivity_agreement)
* [`test_agreement_termination`](https://github.com/golemfactory/yapapi/tree/master/tests/goth/test_agreement_termination)

It would be usefult to port these test cases to `yajsapi`.

yajsapi/executor/agreements_pool.ts
```diff
@@ -0,0 +1,192 @@
...
+    this._cancellation_token = cancellation_token;
+  }
+  async cycle(): Promise<void> {
+    for (const agreement_id of new Map(this._agreements).keys()) {
+      let buffered_agreement = this._agreements.get(agreement_id);
+      if (buffered_agreement === undefined) { continue; }
+      let task = buffered_agreement.worker_task;
+      if (task && task.isFulfilled()) {
+        /* reuse agreement if possible; workers that failed call this method with allow_reuse set to false */
+        await this.release_agreement(buffered_agreement.agreement.id(), true);
```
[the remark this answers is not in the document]
azawlocki: Sorry, I didn't find the second call to `this.release_agreeement()` with `false` since the changes in `index.ts` didn't load automatically.

yajsapi/executor/index.ts
filipgolem: Fixed (`master` branch was not up to date).

yajsapi/executor/agreements_pool.ts
filipgolem: But then `agreement` is not visible in `<code>`, so it's necessary to use `let` instead of `const` and put it above the first `try`.

filipgolem:
> `yapapi` has two integration test cases that test features related to changes in this PR:
> 
> * [`test_multiactivity_agreement`](https://github.com/golemfactory/yapapi/blob/master/tests/goth/test_multiactivity_agreement)
> * [`test_agreement_termination`](https://github.com/golemfactory/yapapi/tree/master/tests/goth/test_agreement_termination)
> 
> It would be usefult to port these test cases to `yajsapi`.

Good idea! I just ported both tests and they are now a part of this PR.

azawlocki (approved):
Looks good, didn't run! Thanks for addressing my comments

PR #173, opened 2021-04-13
Fix map indexing bug

filipgolem:


azawlocki (approved):
How about other uses of `Map` throughout the project? Are there any?

filipgolem:
> How about other uses of `Map` throughout the project? Are there any?

I checked all uses of `Map` and `Set` and it seems there are no other problems like this.

PR #177, opened 2021-04-30
Port subscription expiration from yapapi

filipgolem:
https://github.com/golemfactory/yajsapi/pull/177/commits/5ca13671f7391724f5a4683a8005d31b053220d2 contains all important changes, please review it (a separate commit modifies code indentation).

### DIFF
--- a/yajsapi/executor/index.ts
+++ b/yajsapi/executor/index.ts
@@ -626,7 +626,7 @@ export class Executor implements ComputationHistory {
               if (self._worker_cancellation_token.cancelled) { return; }
               const _batch_timeout = batch.timeout();
               const batch_deadline = _batch_timeout 
-                ? dayjs.utc().unix() + _batch_timeout
+                ? dayjs.utc().unix() + _batch_timeout / 1000
                 : null;
               try {
                 let current_worker_task = consumer.last_item();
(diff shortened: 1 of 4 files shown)


yajsapi/executor/index.ts
```diff
@@ -419,90 +419,104 @@ export class Executor implements ComputationHistory {
...
+        _proposals = subscription.events(self._worker_cancellation_token);
       } catch (error) {
-        emit(new events.SubscriptionFailed({ reason: error }));
-        throw error;
+        emit(
+          new events.CollectFailed({
+            sub_id: subscription.id(),
+            reason: error,
+          })
+        );
```
azawlocki, line 433: Shouldn't we exit this function when `subscription.events()` fails? That's what the `yapapi` version does.

azawlocki (changes requested):
I've found one difference in behaviour between the `yajsapi` and `yapapi` versions of `find_offers_for_subscription()` (see my other comment).

Other than that, `yajsapi` version is about twice as long as the `yapapi` one and has many nested `try ... catch` constructions making it rather hard to follow. I believe it works as it should, but I'm not sure.

azawlocki (approved):
Thanks for replying to my comment

azawlocki (approved):
LGTM!

PR #181, opened 2021-05-10
Repeat API calls on timeouts

filipgolem:


yajsapi/rest/common.ts
```diff
@@ -0,0 +1,55 @@
...
+  block: Callable<void, any>,
+  max_tries: number = 5,
+  max_duration_ms = 15000,
+  interval: number = 0.0,
+  condition: Callable<Error, boolean> = is_intermittent_error
+) {
+  let start_time = Date.now();
+  for (let try_num = 1; try_num <= max_tries; ++try_num) {
+    if (try_num > 1) {
+      await sleep(interval);
```
azawlocki, line 36: Shouldn't we sleep for `min(interval, start_time + max_duration_ms - Date.now())` here?

filipgolem: Yes, then the `max_duration_ms` name would be more relevant.

filipgolem: Fixed

azawlocki (approved):

PR #188, opened 2021-05-18
Return batch results with optional blocking

filipgolem:
- [x] check timeout units (s/ms)
- [x] ctrl+c in `process_batches`
- [x] test `yield`ing pairs from the worker function
- [x] test `wait_for_results: false`.
- [x] smart queue fixes (`yapapi`: #394)
- [x] pass command errors to worker functions (also #394)


azawlocki (approved):
As far as I can tell, looks good!

PR #195, opened 2021-05-20
Change quotes to backticks and set interval between repeated calls to 1s

filipgolem:


zakaprov (approved):

yajsapi/rest/common.ts
```diff
@@ -28,7 +28,7 @@ export async function repeat_on_error(
   block: Callable<void, any>,
   max_tries: number = 5,
   max_duration_ms = 15000,
-  interval: number = 0.0,
+  interval: number = 1000,
```
azawlocki, line 31: Maybe rename this to `interval_ms`? Since the original value was `0.0`, I'd guess that `interval` is in seconde, not ms.

yajsapi/rest/common.ts
```diff
@@ -49,7 +49,7 @@ export async function repeat_on_error(
     const duration = Date.now() - start_time;
     const repeat = try_num < max_tries && duration < max_duration_ms;
     const msg = `API call timed out (attempt ${try_num}/${max_tries}), ` +
-              (repeat ? `retrying in ${interval}s` : "giving up after ${duration}ms");
+              (repeat ? `retrying in ${interval}s` : `giving up after ${duration}ms`);
```
azawlocki, line 52: That will print `retrying in 1000s`!

yajsapi/rest/common.ts
filipgolem: Fixed

yajsapi/rest/common.ts
filipgolem: Fixed

PR #201, opened 2021-05-31
Add stdout/stderr support in CommandEventContext

filipgolem:


yajsapi/executor/index.ts
```diff
@@ -598,7 +598,7 @@ export class Executor {
             emit(evt);
             results.push(evt);
             if (evt instanceof events.CommandExecuted && !evt.success) {
-              throw new CommandExecutionError(evt.command, evt.message)
+              throw new CommandExecutionError(evt.command, evt.stderr)
```
azawlocki, line 601: In `yapapi`, `CommandExecutionError` now has both `message` and `stderr` fields. You may want to change it also in the JS version.

azawlocki: If not, I'd suggest leaving `evt.message` as the second argument of the constructor.

azawlocki (changes requested):

yajsapi/executor/index.ts

azawlocki (approved):
LGTM!

PR #211, opened 2021-06-10
Fix cancellation tokens when submit() is called multiple times.

filipgolem:


yajsapi/executor/index.ts
```diff
@@ -644,7 +637,7 @@ export class Executor {
 
       emit(new events.WorkerStarted({ agr_id: agreement.id() }));
 
-      if (self._worker_cancellation_token.cancelled) { return; }
+      if (self.state && self.state.worker_cancellation_token.cancelled) { return; }
```
azawlocki, line 640: I'd add some kind of type casting, to tell the compiler that `self.state` is expected to be defined at this point. Otherwise it may be a bit confusing for the reader.

azawlocki: The same applies to several other places when `self.state.worker_cancellation_token` appears

azawlocki (approved):
LGTM! I've left one minor remark

yajsapi/executor/index.ts

PR #229, opened 2021-06-28
Send valid termination reason

filipgolem:


azawlocki (changes requested):
The changes look good but the integration tests failed.

The error doesn't seem to be related to the changes though -- could you restart the tests?

azawlocki (approved):

PR #236, opened 2021-07-07
Fix DemandBuilder.constraints()

filipgolem:


shadeofblue (approved):
sure, though, this is one of those cases where a comprehensive set of unit tests would be in order :)

filipgolem:
> sure, though, this is one of those cases where a comprehensive set of unit tests would be in order :)

Added some unit tests.

PR #244, opened 2021-07-21
Add support for driver and network in the yacat example

filipgolem:


filipgolem:
Included in another PR

PR #260, opened 2021-09-23
Change devnet and env var

filipgolem:


shadeofblue:
we don't want these changes in 0.4

PR #276, opened 2021-12-20
add Dockerfile

figurestudios:
as there is one in the [yapapi equivalent](https://github.com/golemfactory/yapapi/blob/master/examples/hello-world/Dockerfile) of this repo, i figured there should be one for this one too

### DIFF
--- a/examples/hello-world/Dockerfile
+++ b/examples/hello-world/Dockerfile
@@ -1,0 +1,3 @@
+FROM alpine:latest
+VOLUME /golem/input
+WORKDIR /golem/work


figurestudios:
I also see that the Task Example mentions ./hello.py, when there is only a [./hello.js](https://github.com/golemfactory/yajsapi/tree/master/examples/hello-world) file in this directory. This is after changing it to view the NodeJS examples:
image
image

filipgolem (approved):
Image hash is the same here and in `yapapi` (`d646d7b93083d817846c2ae5c62c72ca0507782385a2e29291a3d376`), so this Dockerfile is correct.

PR #279, opened 2022-01-28
Fix Winston log level type

filipgolem:


johny-b (approved):
I don't know typescript, but looks OK :)

PR #283, opened 2022-02-21
Update goth to version 0.10+

zakaprov:
This is to prevent an issue with one of `goth`'s transitive dependencies, a similar issue is decribed here: https://github.com/aws/aws-sam-cli/issues/3661

This PR also includes some Actions workflow improvements:
- integration test reruns for some specific exceptions using `pytest-rerunfailures`
- fix Actions job responsible for getting the latest stable branch name


shadeofblue (approved):
looks good to me 👍

PR #289, opened 2022-03-22
287 vpn poc

mgordel:


yajsapi/executor/index.ts
```diff
@@ -171,18 +175,22 @@ export class Executor {
...
    * @param max_workers     maximum number of workers doing the computation
    * @param timeout         timeout for the whole computation
    * @param budget          maximum budget for payments
    * @param strategy        market strategy used to select providers from the market (e.g. LeastExpensiveLinearPayuMS or DummyMS)
    * @param subnet_tag      use only providers in the subnet with the subnet_tag name (env variable equivalent: YAGNA_SUBNET)
-   * @param driver          name of the payment driver to use or null to use the default driver; only payment platforms with the specified driver will be used (env variable equivalent: YAGNA_PAYMENT_DRIVER)
-   * @param network         name of the network to use or null to use the default network; only payment platforms with the specified network will be used (env variable equivalent: YAGNA_PAYMENT_NETWORK)
+   * @param payment_driver  name of the payment driver to use or null to use the default driver; only payment platforms with the specified driver will be used (env variable equivalent: YAGNA_PAYMENT_DRIVER)
```
shadeofblue, line 190: I believe we should still leave the old arguments in place and add a deprecation warning when they're used ... 

and only remove the old arguments in a subsequent minor release

shadeofblue:
nice 👍  

I think it's nice from consistency point of view that you have decided to mimic the Python interface 👍 

please have a look at my comment re: deprecation process

yajsapi/executor/index.ts
mgordel: Yes, you are absolutely right. Too quickly I refactored the name to a more meaningful, forgetting about backwards compatibility.

yajsapi/executor/index.ts
```diff
@@ -194,12 +202,13 @@ export class Executor {
     driver,
     network,
```
zakaprov, line 203: I'm aware this would change the public API, but maybe we should update the parameter names here as well (like we did for internal `Executor` fields)?

yajsapi/executor/index.ts
```diff
@@ -945,8 +964,21 @@ export class Executor {
...
     let payment_client = await this._api_config.payment();
     this._payment_api = new rest.Payment(payment_client);
+
+    let net_client = await this._api_config.net();
+    this._net_api = new rest.Net(net_client);
+
     await stack.enter_async_context(this._wrapped_consumer);
 
+    if (this._network_address) {
+      // TODO: replace with a proper REST API client once ya-client and ya-ts-client are updated
```
zakaprov, line 974: Let's create a GitHub issue and add it to the SDK project to track this!

tests/mock/rest.ts
```diff
@@ -0,0 +1,20 @@
...
+    super(null);
+  }
+  setExpected(function_name, result, error?) {
+    this[function_name] = async () => new Promise((res, rej) => {
+      if (error) rej(error);
+      res(result);
+      result = 'test';
+    })
+  }
+  async create_network(ip: string, mask?: string, gateway?: string): Promise<string> { return 'test'; }
```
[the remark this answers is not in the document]
mgordel: 👍 as Uncle Bob says: meaningful names matter

yajsapi/executor/index.ts
mgordel: I added new param names and left the old ones with a deprecation warning

examples/ssh/ssh.js
```diff
@@ -0,0 +1,67 @@
+const { Executor, Task, utils, vm } = require("../../dist");
+const { asyncWith, sleep } = utils;
+const crypto = require("crypto");
+const { program } = require("commander");
+
+async function main(subnetTag, driver, network, count = 2, session_timeout = 60) {
+  const _package = await vm.repo({
+    image_hash: "1e06505997e8bd1b9e1a00bd10d255fc6a390905e4d6840a22a79902",
```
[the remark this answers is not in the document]
mgordel: comment added

yajsapi/executor/index.ts
mgordel: https://github.com/golemfactory/yajsapi/issues/290

zakaprov (approved):
Awesome! Thanks a lot for implementing the suggestions. :)

Issue #293, opened 2022-04-07
Cleanup repo


mgordel:
- [x] nodejs update
- [x] update old dependencies
- [x] linter rules
- [x] tests
- [x] docs
- [x] yarn warnings - #204

mgordel:
- [ ] nodejs version
- [ ] update old dependencies
- [ ] linter rules
- [ ] test framework

mgordel:
TODO:
 - `gftp_test.ts` and `unsubscribe.ts` in examples..
 - `sgx-ias-js` outdated package - conflict with `dayjs`

PR #294, opened 2022-04-08
Updated Nodejs. Updated Typescript. Added ESLint. Added Prettier. Fix…

mgordel:
Updates and cleanup (#293)


README.md
```diff
@@ -5,7 +5,7 @@ The JS/TS port of [yapapi, Golem's Python API](https://github.com/golemfactory/y
 
 ## Prerequisites
 
-- [node](https://nodejs.org/en/) >= v12.13.0
+- [node](https://nodejs.org/en/) >= v14.17.0
```
shadeofblue, line 8: https://github.com/golemfactory/yagna-docs/issues/110

examples/hello-world/hello.js
```diff
@@ -11,18 +16,15 @@ async function main() {
...
     }
   }
 
-  await asyncWith(
-    new Executor({ task_package: package, budget: "1.0", subnet_tag: "devnet-beta" }),
-    async (executor) => {
-      for await (let completed of executor.submit(worker, tasks)) {
-        console.log(completed.result().stdout);
-      }
+  await asyncWith(new Executor({ task_package, budget: "1.0", subnet_tag: "devnet-beta" }), async (executor) => {
```
shadeofblue, line 23: actually, I'd recommend against this kind of change, in my opinion it decreases the code's readability...

examples/low-level-api/listOffers.ts
```diff
@@ -2,7 +2,7 @@ import Bluebird from "bluebird";
 import dayjs from "dayjs";
 import utc from "dayjs/plugin/utc";
 import { props as yp, rest, utils } from "yajsapi";
-import { Subscription } from "yajsapi/dist/rest/market";
+import { Subscription } from "../../yajsapi/rest/market";
```
shadeofblue, line 5: hmm... doesn't it mean that we're depending on the source version here? I guess it would be better to use the installed npm package if possible...

examples/low-level-api/listOffers.ts
```diff
@@ -12,22 +12,22 @@ const { asyncWith, CancellationToken } = utils;
 const cancellationToken = new CancellationToken();
 
 async function list_offers(conf: rest.Configuration, subnetTag: string): Promise<void> {
-  let client = await conf.market();
-  let market_api = new Market(client);
-  let dbuild = new yp.DemandBuilder();
+  const client = await conf.market();
+  const market_api = new Market(client);
+  const dbuild = new yp.DemandBuilder();
```
shadeofblue, line 17: okay, here's where my js is not sharp enough -> doesn't `const` suggest that the variable will not be modified? I guess the meaning is obvious when it comes to immutable values... but here it's a mutable and while indeed, we won't be assigning to the `dbuild`, I guess its value _will_ get modified, won't it? ... 

that doesn't feel like a constant to me then... but if it's still okay by the js lingo, I'm okay

examples/unsubscribe.ts
```diff
@@ -1,16 +1,16 @@
 import dayjs from "dayjs";
 import utc from "dayjs/plugin/utc";
 
-import { rest } from "yajsapi";
+import { rest } from "../yajsapi";
```
shadeofblue, line 4: again, doesn't it mean that we're depending on the source version of yajsapi?

tests/goth/test_run_blender.py
```diff
@@ -103,7 +103,7 @@ async def test_run_blender(
         requestor = runner.get_probes(probe_type=RequestorProbe)[0]
 
         async with requestor.run_command_on_host(
-            f"node {blender_path} --subnet-tag goth",
+            f"npm run --prefix {examples_dir} ts:blender -- -d --subnet-tag goth",
```
shadeofblue, line 106: just out of curiosity - what's the difference between these two invocations?

tests/utils/promisify.ts
```diff
@@ -18,5 +18,5 @@ test("promisify reject given fn", async (t) => {
   }
 
   const error = await t.throwsAsync(promisify(wantCb)(21));
-  t.is(error.message, "i am an error");
+  t.is(error!.message, "i am an error");
```
shadeofblue, line 21: could you explain this syntax? (again, I'm mostly curious)

yajsapi/executor/ctx.ts
```diff
@@ -20,55 +22,59 @@ export class CommandContainer {
...
           }
-          let idx = self._commands.length;
+          const idx = self._commands.length;
           self._commands.push({ [name]: _arguments });
           return idx;
         };
         return new Proxy(newFunction, {
           apply: function (target, thisArg, argumentsList) {
-            return target.apply(thisArg, argumentsList);
+            return target.apply(thisArg, argumentsList as []);
```
shadeofblue, line 44: is this a ts's way of saying "argumentsList" is an array?

yajsapi/executor/ctx.ts
```diff
@@ -233,11 +238,9 @@ class _RecvFile extends Work {
 
   async post(): Promise<void> {
     if (!this._dst_slot) throw "_RecvFile post without prepare";
-    if (this._emitter)
-      this._emitter(new events.DownloadStarted({ path: this._src_path }));
+    if (this._emitter) this._emitter(new events.DownloadStarted({ path: this._src_path }));
```
shadeofblue, line 241: wouldn't this be more readable: 
```suggestion
    if (this._emitter) { 
        this._emitter(new events.DownloadStarted({ path: this._src_path }));
    }
```
?

yajsapi/executor/ctx.ts
```diff
@@ -233,11 +238,9 @@ class _RecvFile extends Work {
...
     await this._dst_slot.download_file(this._dst_path);
-    if (this._emitter)
-      this._emitter(new events.DownloadFinished({ path: this._dst_path }));
+    if (this._emitter) this._emitter(new events.DownloadFinished({ path: this._dst_path }));
```
shadeofblue, line 243: same here...

yajsapi/executor/ctx.ts
```diff
@@ -293,13 +296,13 @@ export class WorkContext {
   private _started: boolean;
   private _emitter: Callable<[StorageEvent], void> | null;
   private _network_node?: NetworkNode;
-  public provider_info: { provider_id: string, provider_name: string };
+  public provider_info: { provider_id: string; provider_name: string };
```
shadeofblue, line 299: hmm.... can you separate dictionary values with semicolons indeed?

yajsapi/executor/index.ts
```diff
@@ -208,40 +205,54 @@ export class Executor {
     payment_driver,
     payment_network,
     event_consumer,
-    network_address
+    network_address,
   }: ExecutorOpts) {
     this._subnet = subnet_tag ? subnet_tag : DEFAULT_SUBNET;
-    this._payment_driver = payment_driver ? payment_driver.toLowerCase() : (driver ? driver.toLowerCase() : DEFAULT_DRIVER);
-    this._payment_network = payment_network ? payment_network.toLowerCase() : (network ? network.toLowerCase() : DEFAULT_NETWORK);
+    this._payment_driver = payment_driver
```
shadeofblue, line 211: hmm... if we're splitting that into multiple lines, wouldn't a regular if/else syntax make more sense?

shadeofblue: plus, then, the deprecation warning could be included in the if/else and so overall, it would result in even less condition branches...

yajsapi/executor/index.ts
```diff
@@ -589,28 +581,28 @@ export class Executor {
...
-              "Overriding batch timeout set with commit({timeout:...}) by value from ExecOptions"
-            );
+            logger.warn("Overriding batch timeout set with commit({timeout:...}) by value from ExecOptions");
           } else {
             exec_options.batch_timeout = batch.timeout();
           }
         }
-        const batch_deadline
-          = exec_options.batch_timeout ? ((Date.now() + exec_options.batch_timeout) / 1000) : undefined;
+        const batch_deadline = exec_options.batch_timeout
```
shadeofblue, line 594: again, if we're splitting the ternary into multiple lines, shouldn't we replace it with a more conventional if/else?

yajsapi/executor/strategy.ts
```diff
@@ -132,7 +115,7 @@ export class LeastExpensiveLinearPayuMS {
 
     // The higher the expected price value, the lower the score.
     // The score is always lower than SCORE_TRUSTED and is always higher than 0.
-    const score: number = (SCORE_TRUSTED * 1.0) / (expected_price + 1.01);
+    const score: number = SCORE_TRUSTED / (expected_price + 1.01);
```
shadeofblue, line 118: isn't the cast to float needed here?

yajsapi/network/index.ts
```diff
@@ -21,6 +29,11 @@ export class NetworkNode {
     };
   }
 
+  /**
```
shadeofblue, line 32: just a comment: I guess docstrings et al could go into a separate pull request...

yajsapi/utils/promisify.ts
```diff
@@ -1,4 +1,4 @@
-export default function promisify(fn: Function): Function {
+export default function promisify(fn: (...args) => void): (arg) => Promise<any> {
```
shadeofblue, line 1: I think I don't get the type definition for `fn` here -> doesn't it limit the functions passed here to ones that don't return any values? is that what is wanted?

shadeofblue (approved):
I guess I didn't find anything really out of place here ... but given the length of the pull request, I may have missed something... 

otherwise, I added some comments and questions...

examples/hello-world/hello.js
mgordel: this kind of formatting has been corrected automatically by the `prettier` tool with default rules. This can be tweaked of course, but I would leave the default rules which are often used in the js community

examples/low-level-api/listOffers.ts
mgordel: this kind of formatting has been corrected automatically by the `prettier` tool with default rules. This can be tweaked of course, but I would leave the default rules which are often used in the js community

examples/low-level-api/listOffers.ts
mgordel: yeah .. JS can be weird sometimes. In JS, `const` means that the identifier can’t be assigned, not that values is immutable. And linter show errors if you use `let` for identifiers which are not reassigned.

examples/unsubscribe.ts
mgordel: yes, you are right, i'll fix it.

tests/goth/test_run_blender.py
mgordel: in the first one we use node directly so we can only execute js file. we can't run ts file (typescript) in node engine. First we need to compile ts -> js. To avoid this in this test example (i wanted to test ts version in goth), we can use the ts-node tool, which compiles and runs the ts file on the fly. It is hidden under the `ts:blender` command in package.json, so if we want to run more advanced scripts than node {js_file} we need to define it in package.json and use npm, something like poetry and poe tool..

tests/utils/promisify.ts
mgordel: this is TS syntax. The error variable can be undefined (by default type is Error | undefined), in which case referring to the message property will throw error. If we don't define an if statement that checks if error is undefined, the typescript compiler will report an error. The syntax `error!.message` is such a hack that tells TS compiler that we know that error var will definitely not be undefined, and ts compiler not show error in this case.

yajsapi/executor/ctx.ts
mgordel: Yes exactly

yajsapi/executor/ctx.ts
mgordel: Yes, i agree

yajsapi/executor/ctx.ts
mgordel: ts suggests using `type` instead of an object (dictionary).
the `type` construct uses semicolons. https://www.typescriptlang.org/docs/handbook/2/objects.html

yajsapi/executor/index.ts
mgordel: yes, i agree

yajsapi/utils/promisify.ts
mgordel: yes, it should be `(...args) => any`

examples/low-level-api/listOffers.ts
mgordel: yes, you are right, i'll fix it.

PR #297, opened 2022-04-19
Fix "number is not assignable to bigint" error

filipgolem:


mgordel (approved):
dependency ip-num was updated 3 days ago
https://github.com/ip-num/ip-num/releases/tag/v1.4.0

PR #298, opened 2022-04-19
291 replace asyncwith

mgordel:


examples/package.json
```diff
@@ -26,7 +26,7 @@
         "commander": "^9.1.0",
         "dayjs": "~1.9.1",
         "ts-node": "^10.7.0",
-        "yajsapi": "^0.6.0"
+        "yajsapi": "file:../dist"
```
zakaprov, line 29: Should this be updated before merging?

zakaprov (approved):
:ok:

examples/package.json
mgordel: Yes, my mistake

shadeofblue:
one (maybe important) remark that I have is that if we're anyway removing the `asyncWith` and requiring the Executor instance to be created before we're able to use it, maybe we could also provide a start/stop interface where we don't have to limit what we do with that intance to the body of the function passed to `run` ... 

but maybe that can be done separately...

shadeofblue (approved):
besides the comment above, it looks good

PR #309, opened 2022-06-08
bump goth to 0.11

shadeofblue:


.github/workflows/unit-tests.yml
```diff
@@ -15,7 +15,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [12.x, 14.x, 15.x]
+        node-version: [14.x, 16.x, 17.x]
```
mgordel, line 18: In the b06 branch there were no changes related to the nodejs upgrade. The tests should pass for newer versions, but since we support nodejs 12 in this branch, maybe it's better to leave it?

mgordel (approved):

.github/workflows/unit-tests.yml
shadeofblue: yeah, but I think that github stopped supporting it ? ...

PR #313, opened 2022-06-20
Integrate new mid-level activity module (#308)

mgordel:


yajsapi/executor/index.ts
```diff
@@ -17,7 +17,9 @@ import * as rest from "../rest";
 import { Agreement, OfferProposal, Subscription } from "../rest/market";
 import { AgreementsPool } from "./agreements_pool";
 import { Allocation, DebitNote, Invoice } from "../rest/payment";
-import { ActivityService, CommandExecutionError } from "../rest/activity";
+import { CommandExecutionError } from "../rest/activity";
+
+import { ActivityFactory, Activity } from "../mid-level-api/activity";
```
shadeofblue, line 22: I'd still prefer not to have the "mid-level-api" named as a specific module -> it's, to an extent, artificial and non-functional ... I'd just put this `activity` module in the same directory tree as other specialized modules, that is, as: `yajsapi/activity`... 

I'm not pushing for this to happen as part of this pull request though, just wouldn't like for this to remain in place after we finalize the whole "activity" saga

yajsapi/executor/index.ts
```diff
@@ -50,6 +52,8 @@ import {
 } from "./strategy";
 import { Package } from "../package";
 import axios from "axios";
+import { Script, Command } from "../mid-level-api";
```
shadeofblue, line 55: again, this might be `yajsapi/script` or `yajsapi/activity/script`

yajsapi/executor/index.ts
```diff
@@ -600,26 +604,46 @@ export class Executor {
...
+
         const get_batch_results = async (): Promise<events.CommandEvent[]> => {
           const results: events.CommandEvent[] = [];
-          for await (const evt_ctx of remote) {
-            const evt = evt_ctx.event(agreement_id, task_id, cmds);
+          scriptResults.on("error", (error) => {
+            throw new CommandExecutionError("", error.toString());
+          });
+          for await (const result of scriptResults) {
+            let evt = Object.create(events.CommandExecuted.prototype);
```
shadeofblue, line 638: please extract the creation of this event somewhere (either to the workcontext or to the activity module) instead of coding it inline here...

shadeofblue:
@mgordel while the change of the location of the activity module (so as not to have "mid-level-api" as a module), might be done in a separate pull request indeed, I'd like to have this comment addressed: https://github.com/golemfactory/yajsapi/pull/313#discussion_r913431227

shadeofblue (changes requested):
@mgordel while the change of the location of the activity module (so as not to have "mid-level-api" as a module), might be done in a separate pull request indeed, I'd like to have this comment addressed: https://github.com/golemfactory/yajsapi/pull/313#discussion_r913431227

PR #314, opened 2022-06-20
Added script builder and adapter for new storage provider (#308) 

mgordel:


yajsapi/mid-level-api/script/command.ts
```diff
@@ -0,0 +1,95 @@
...
+    this.args["from"] = await this.storageProvider.upload(this.srcPath);
+  }
+}
+
+export class DownloadFile extends Transfer {
+  constructor(private storageProvider: StorageProvider, private srcPath: string, private dstPath: string) {
+    super();
+    this.args = { from: `container:${srcPath}` };
+  }
+  async before() {
```
shadeofblue, line 92: @mgordel why does the download happen in `before`? the file has first to be transferred from the runtime to GFTP and only downloaded in `after`? am I wrong?

shadeofblue: have you tested these transfer commands both ways? (upload and download)?

yajsapi/mid-level-api/script/command.ts
```diff
@@ -0,0 +1,95 @@
...
+  }
+}
+
+export class Transfer extends Command {
+  constructor(protected from?: string, protected to?: string, args?: object) {
+    super("transfer", { from, to, args });
+  }
+}
+
+export class SendFile extends Transfer {
```
shadeofblue, line 77: @mgordel I'd change the name here to `UploadFile` _or_ change `DownloadFile` to `ReceiveFile`

iow, either we're using upload+download or send+receive ...

yajsapi/mid-level-api/script/command.ts
shadeofblue: okay, the name `download` confused me, as discussed, those methods should probably be renamed to `publish` vs `receive`

shadeofblue (changes requested):
please see my comments

yajsapi/mid-level-api/script/command.ts
mgordel: ok. I changed to UploadFile

yajsapi/mid-level-api/script/command.ts
mgordel: I change them to publish and receive

PR #341, opened 2022-08-12
Fixes for examples without yajsapi backward compatibility (#340)

mgordel:


VanDavv:
Gave a hello world example a try on my Debian 11 and works great now!
Will also give it a try on Windows later today, but I think LGTM

VanDavv (approved):

VanDavv (approved):

PR #345, opened 2022-08-24
344 manifest support

mgordel:


mgordel:
This is already obsolete and does not match the new code

PR #364, opened 2022-09-20
355 taks api new features description markdown

### LINKED ISSUE #355: Taks API - new features described in .md file
as developer I get .md file that contains description of new functions in the Task API and examples

mgordel:
closes #355

### DIFF
--- a/tutorial.md
+++ b/tutorial.md
@@ -1,0 +1,33 @@
+# New Task API
+
+The new version of the Taks API introduces a simplified interface for executing commands in the Golem network.
+
+### The basic algorithm is as follows:
+
+1. Creating a Executor instance
+
+    The executor can be created by passing appropriate initial parameters such as: ...
+    One required parameter is image_hash.
+
+   2. Execute one of the available executor methods: `run`, `map` or `foreach`.
+
+       a) Method `run` - we use method run to execute one worker function, eg.
+       ```js
+           const result = await executor.run((ctx) => ctx.run("echo 'Hello World!'"));
+       ```
+   
+       b) Method `map` - we use method map to get results for each element in iterable object, eg,
+       ```js
+           const data = [1, 2, 3, 4, 5];
+           const results = executor.map(data, (ctx, item) => ctx.run(`echo "${item}"`));
+           for await (const result of results) console.log(result);
+       ```
+       In results we get async iterable object, so we can iterate fo each element by `for await` function
+
+3. Termination an executor instance
+
+   Termination of contracts, payment processing, etc.
+
+### Worker Function and Work Context API
+
+    TODO
--- a/tutorial.md
+++ b/tutorial.md
@@ -80,7 +80,7 @@ const res = await ctx
 
 res?.map(({ stdout }) => console.log(stdout));
 ```
-You can end batch of command by `end()` as above and it means that this code return `Promise` of `Result` objects (or throw an error if occurred).
+You can end the batch of commands by using the `end()` method shown above, meaning this code returns a `Promise` of multiple `Result` objects (or throw an error if occurred).
 
 You can also end this batch by `endStream()` to get a `Readable` stream, e.g.
 
--- a/tutorial.md
+++ b/tutorial.md
@@ -4,7 +4,7 @@ The new version of the Task API introduces a simplified interface for executing 
 
 ## The basic algorithm
 
-### 1. Creating a Executor instance
+### 1. Creating an Executor instance
 
 The executor can be created by passing appropriate initial parameters such as package, budget, subnet tag, payment driver, payment network etc.
 One required parameter is a package. This can be done in two ways. First by passing only package image hash, e.g.
--- a/tutorial.md
+++ b/tutorial.md
@@ -2,32 +2,99 @@
 
 The new version of the Taks API introduces a simplified interface for executing commands in the Golem network.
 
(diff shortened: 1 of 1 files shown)


tutorial.md
```diff
@@ -0,0 +1,117 @@
...
+  console.log(result.stdout)
+```
+
+#### b) `map` - to get results for each element in iterable object, e.g.
+```js
+  const data = [1, 2, 3, 4, 5];
+  const results = executor.map(data, (ctx, item) => ctx.run(`echo "${item}"`));
+  for await (const result of results) console.log(result.stdout);
+```
+In results we get async iterable object, so we can iterate for each element by `for await` statement.
```
VanDavv, line 38:
```suggestion
In the `results` variable, we have an async iterable object, with each element accessible with the `for await` statement.
```

tutorial.md
```diff
@@ -0,0 +1,117 @@
...
+
+#### c) `forEach` method - it is very similar to map, but it does not return any value, e.g.
```
VanDavv, line 40:
```suggestion
#### c) `forEach` method - it is very similar to a map, but it does not return any value, e.g.
```

tutorial.md
```diff
@@ -0,0 +1,117 @@
...
+```js
+  const data = [1, 2, 3, 4, 5];
+  await executor.forEach(data, async (ctx, item) => {
+      console.log((await ctx.run(`echo "${item}"`).stdout));
+  });
+```
+
+### 3. Termination an executor instance
```
VanDavv, line 48:
```suggestion
### 3. Termination of an executor instance
```

tutorial.md
```diff
@@ -0,0 +1,117 @@
...
+Termination of contracts, payment processing, etc.
+
+```js
+await executor.end();
+```
+
+
+### Worker Function and Work Context API
+
+Each of available method: `run`, `map` and `forEach` takes as a parameter `worker` function. Worker function is asynchronous and provide `Work Contect API` - `ctx` object.
```
VanDavv, line 59:
```suggestion
Each of the available methods: `run`, `map` and `forEach` takes the `worker` function as a parameter. The worker function is asynchronous and provides a `Work Context API` - `ctx` object.
```

tutorial.md
```diff
@@ -0,0 +1,117 @@
...
+
+Work Context allow to run single commands or batch of commands on provider.
```
VanDavv, line 61:
```suggestion
Work Context allows running single commands or batches of commands on a provider.
```

tutorial.md
```diff
@@ -0,0 +1,117 @@
...
+
+Single commands available to run on provider:
```
VanDavv, line 63:
```suggestion
Single commands available to run on a provider:
```

tutorial.md
```diff
@@ -0,0 +1,117 @@
...
+
+   - `run()`
+   - `uploadFile()`
+   - `uploadJson()`
+   - `downloadFile()`
+
+You can also compose particular commands into batch. For this you should use `beginBatch`, e.g.
```
VanDavv, line 70:
```suggestion
You can also compose particular commands into batches. For this, you should use `beginBatch`, e.g.
```

tutorial.md
```diff
@@ -0,0 +1,117 @@
...
+   .beginBatch()
+   .run('echo "Hello Golem"')
+   .run('echo "Hello World"')
+   .uploadJson({ hello: 'Golem'}, '/golem.json')
+   .end()
+   .catch((error) => console.error(error));
+
+res?.map(({ stdout }) => console.log(stdout));
+```
+You can end batch of command by `end()` as above and it means that this code return `Promise` of `Result` objects (or throw an error if occurred).
```
VanDavv, line 83:
```suggestion
You can end the batch of commands by using the `end()` method shown above, meaning this code returns a `Promise` of multiple `Result` objects (or throw an error if occurred).
```

tutorial.md
```diff
@@ -0,0 +1,117 @@
...
+
+You can also end this batch by `endStream()` and then you get `Readable` stream, e.g.
```
VanDavv, line 85:
```suggestion
You can also end this batch by `endStream()` to get a `Readable` stream, e.g.
```

tutorial.md
```diff
@@ -0,0 +1,117 @@
...
+   .endStream();
+
+ results.on("data", ({ stdout }) => console.log(stdout));
+ results.on("error", (error) => console.error(error));
+ results.on("close", () => console.log("END"));
+```
+
+### Additional executor method `beforeEach()` 
+
+There is special method available in executor object, that allow you to run the worker function once before each workers on provider per one activity. This is `beforeEach()` method.
```
VanDavv, line 102:
```suggestion
There is a special method available in the executor object, that allows you to run the worker function once before each worker executes other tasks on the provider, but within the same activity. This is `beforeEach()` method.
```

tutorial.md
```diff
@@ -0,0 +1,117 @@
...
+This method takes as parameter one worker function and run it only once for each new provider activity. The following example demonstrate the use of this method.
```
VanDavv, line 103:
```suggestion
This method takes as a parameter one worker function and runs it only once for each new provider activity. The following example demonstrates the use of this method.
```

VanDavv (changes requested):
Great tutorial we can iterate further, thanks!

PR #372, opened 2022-10-18
370 dns client srv record package resolver url compatible with the browser

Braunmann:


shadeofblue:
hmm... I have a serious concerns about using a very centralized tool such as a google resolver in a purposefully decentralized tool that we're building

Braunmann:
We can use native node support [node.dns](https://nodejs.org/docs/latest/api/dns.html#class-dnsresolver) when resolver is run as node app. 
When it will be run in browser we can use temporary Google DNS API. 
To consider in future we can use our API to provide correct address - but still it will be one entry point - I think this topic TBD

Braunmann:
@mgordel Please review again

mgordel (approved):
I

PR #378, opened 2022-10-19
Logger extracted as dependencies (#396)

mgordel:


Braunmann:
@mgordel  `utils.changeLogLevel("debug");` seems to be not avaliable anymore. I think it will be nice to have same interface .

yajsapi/utils/index.ts
```diff
@@ -6,13 +6,13 @@ import Callable from "./callable";
 import CancellationToken from "./cancellationToken";
 import eventLoop from "./eventLoop";
 import getAllProperties from "./getAllProperties";
-import logger, * as logUtils from "./log";
-import { changeLogLevel } from "./log";
```
Braunmann, line 10: Seems like `utils.changeLogLevel("debug");` is not avaliable anymore - I think it is good idea to do not change interface, or correct examples

mgordel: Yes, you're right. I forgot about examples. Now logger is a param of executor constructor and has method setLevel to change log level. For the default logger (winston for nodejs), when the param is null, we may add logLevel parameter to the executor constructor. I'll fix it.

mgordel: @Braunmann please review again

yajsapi/rest/payment.ts
```diff
@@ -338,9 +338,9 @@ export class Payment {
           "getInvoiceEvents"
         );
         for (const ev of events) {
-          self.logger?.debug(
-            `Received invoice event: ${JSON.stringify(ev)}, ` + `type: ${JSON.stringify(Object.getPrototypeOf(ev))}`
-          );
+          // self.logger?.debug(
```
Braunmann, line 341: Is it consciously commented out?

yajsapi/rest/payment.ts
```diff
@@ -382,9 +382,9 @@ export class Payment {
           "getDebitNoteEvents"
         );
         for (const ev of events) {
-          self.logger?.debug(
-            `Received debit note event: ${JSON.stringify(ev)}, ` + `type: ${JSON.stringify(Object.getPrototypeOf(ev))}`
-          );
+          // self.logger?.debug(
```
Braunmann, line 385: Is it consciously commented out?

mgordel: Yes, temporary I disabled this because it caused bug described in #381

PR #412, opened 2022-11-15
402 agreement pool service new

Braunmann:
Still is not ready for final PR.


yajsapi/agreement/agreement.ts
```diff
@@ -0,0 +1,80 @@
...
+  eventPoolingMaxEventsPerRequest?: number;
+  logger?: Logger;
+}
+
+export interface ProviderInfo {
+  providerName: string;
+  providerId: string | null;
+}
+
+export class Agreement {
```
mgordel, line 22: What if the Agreement had such an interface created by factory?
These are all the things we need right now
```js
class Agreement {
  public readonly id;
  public readonly providerId: string;
  public readonly providerName: string;

  async getState(): Promise<AgreementStateEnum> {}
  async confirm() {}
  async terminate() {}
}
```

yajsapi/agreement/agreement.ts
```diff
@@ -0,0 +1,80 @@
...
+      providerId: this.agreementData?.offer?.providerId || null,
+    };
+  }
+
+  async getState(): Promise<AgreementStateEnum> {
+    await this.refreshDetails();
+    return this.agreementData!.state;
+  }
+
+  async isFinalState(): Promise<boolean> {
```
mgordel, line 52: I wonder if it should not be on the service? (In the context of using mid-level standalone) Only service knows what finalState is... ?

PR #419, opened 2022-12-01
402 agreement pool service new

Braunmann:


mgordel:
It seems ok. We clean the code and make it consistent with other services after merging them.

PR #421, opened 2022-12-05
400 payment service

mgordel:


yajsapi/market/service.ts
```diff
@@ -81,6 +64,15 @@ export class MarketService {
   }
 
   private isProposalValid(proposal: Proposal): { result: boolean; reason?: string } {
+    // TODO: !!!!
```
Braunmann, line 67: Marcin are you aware about TODO?

Braunmann (approved):

PR #423, opened 2022-12-06
Decorations builder refactor

Braunmann:


yajsapi/market/decorations_builder.ts
```diff
@@ -0,0 +1,69 @@
+import { MarketProperty } from "ya-ts-client/dist/ya-payment/src/models/market-property";
+import { MarketDecoration } from "ya-ts-client/dist/ya-payment/src/models/market-decoration";
+
+export enum ComparisonMethod {
```
mgordel, line 4: Maybe `ComparisionOperator` would be a better name..? like in https://www.mongodb.com/docs/manual/reference/operator/query-comparison/

yajsapi/market/decorations_builder.ts
```diff
@@ -0,0 +1,69 @@
+import { MarketProperty } from "ya-ts-client/dist/ya-payment/src/models/market-property";
+import { MarketDecoration } from "ya-ts-client/dist/ya-payment/src/models/market-decoration";
```
mgordel, line 2: it can be shortened
```suggestion
import { MarketProperty, MarketDecoration } from "ya-ts-client/dist/ya-payment/src/models";
```

mgordel (approved):

PR #432, opened 2022-12-28
407 events

Braunmann:


yajsapi/executor/executor.ts
```diff
@@ -101,11 +101,15 @@ export class TaskExecutor {
     await this.paymentService.end();
     this.storageProvider?.close();
     this.options.eventTarget?.dispatchEvent(new Events.ComputationFinished());
-    this.logger?.table?.(this.statsService.getAllCosts());
+    this.logger?.table?.(this.statsService.getStatsTree());
```
mgordel, line 104: This structure is not fit the table

yajsapi/stats/abstract_aggregator.ts
```diff
@@ -0,0 +1,31 @@
+import collect, { Collection } from "collect.js";
```
mgordel, line 1: Is it necessary to use an external library if we only use the map function?

yajsapi/stats/providers.ts
```diff
@@ -1,57 +1,24 @@
...
-    return [...this.agreements].map(([id, data]) => ({ ...data, id }));
-  }
-
-  getProviderName(agreementId: string): string | undefined {
-    return this.agreements.get(agreementId)?.providerName;
+export class Providers extends AbstractAggregator<Payload, ProviderInfo> {
+  beforeAdd(payload): ProviderInfo {
+    if (payload.providerName) {
+      return payload;
+    } else {
```
mgordel, line 16: Else is not necessary if preceded by "return"

yajsapi/stats/service.ts
```diff
@@ -43,60 +52,106 @@ export class StatsService {
...
-        Cost: costs.amount,
-        "Payment Status": costs.paid ? "paid" : "unpaid",
-      };
-    });
-  }
-
-  getComputationsInfo() {
-    // todo
-  }
+  getStatsTree() {
```
mgordel, line 55: we should consider what this structure should look like...

mgordel (approved):

PR #480, opened 2023-03-21
479 gftp storage provider releasing urls

mgordel:


Braunmann (approved):
Reviewed,  tested, the error does not occur anymore

Issue #478, opened 2023-03-20
Logging improvements


mgordel:
Adding configuration options to the PinoLogger so that can save logs to a file in the appropriate form (csv, json??)

pgrzy-golem:
Solution: Use factory functions to initialize loggers, leave default to console logging.

- [ ] Create null logger factory in utils module (to be able to disable the logging completely) 
- [ ] Create a console logger factory in utils module (used by default if no logger is provided)
- [ ] Create a file logger that will write logs to file
- [ ] Create unit tests

PR #481, opened 2023-03-22
Added logger factories producing logger instances for many use-cases …

pgrzy-golem:
…(#478)


Braunmann (approved):

yajsapi/utils/index.ts
mgordel, line 1: I wonder if it can be somehow exported in a prettier form..?
Maybe it would be better to wrap it in a `loggers` object with all things related to the logger, then it would be: `import { loggers } from 'yajsapi'` and `loggers.createDefaultLogger` ... What do you think?

pgrzy-golem: Let's discuss it

PR #487, opened 2023-04-24
485 agreement pool strategy fixes

Braunmann:


tests/mock/services/market_startegy.ts
```diff
@@ -1,11 +1,12 @@
...
 
 export const marketStrategyAlwaysBan: MarketStrategy = {
-  scoreProposal(proposal: Proposal): number {
-    return -1;
+  async checkProposal(proposal: ProposalDTO): Promise<boolean> {
+    return true;
   },
-  getDemandDecoration(): MarketDecoration {
-    return {} as MarketDecoration;
+  async getBestAgreementCandidate(candidates: AgreementCandidate[]): Promise<AgreementCandidate> {
```
pgrzy-golem, line 9: I know we agreed upon this name, but here's just something to consider. "getBestAgreementCandidate" suggests we want to choose the best out of the given set. The thing is we already discussed providing lower scored or even random candidate in order to spread the work across more providers.

So the "best" in the name suggests already some sort of strategy here. I think we should consider another name before we release this to the public.

pgrzy-golem: some suggestions
- getNextAgreementCandidate
- pickAgreementCandidate
- nextAgreementCandidate
- getAgreementCandidate
- chooseAgreementCandidate
- selectAgreementCandidate

pgrzy-golem (approved):

Issue #265, opened 2021-11-15
update examples to use erc20/rinkeby by default


shadeofblue:
* update examples to use `payment_driver=erc20` and `payment_network=rinkeby` by default
* ensure this default is tested and works correctly with `goth`

possibly modify https://github.com/golemfactory/yajsapi/pull/264

PR #492, opened 2023-05-19
Fixed process.env related bug (#265)

pgrzy-golem:
Should solve this: https://github.com/golemfactory/yagna-sdk-team/issues/265

### DIFF
--- a/yajsapi/market/config.ts
+++ b/yajsapi/market/config.ts
@@ -36,7 +36,7 @@ export class DemandConfig {
     const apiConfig = new Configuration({ apiKey, basePath: `${basePath}/market-api/v1`, accessToken: apiKey });
     this.yagnaOptions = options?.yagnaOptions;
     this.api = new RequestorApi(apiConfig);
-    this.subnetTag = options?.subnetTag || (typeof process !== 'undefined' ? process.env.YAGNA_SUBNET : '') || DEFAULTS.subnetTag;
+    this.subnetTag = options?.subnetTag || EnvUtils.getYagnaSubnet() || DEFAULTS.subnetTag;
     this.timeout = options?.marketTimeout || DEFAULTS.marketTimeout;
     this.expiration = options?.marketOfferExpiration || DEFAULTS.marketOfferExpiration;
     this.offerFetchingInterval = options?.offerFetchingInterval || DEFAULTS.offerFetchingInterval;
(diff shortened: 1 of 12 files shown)


yajsapi/utils/env.ts
mgordel, line 1: Perhaps instead of a class (where we do not want a constructor) it is better to use functions? e.g.
```typescript
export function getYagnaApiUrl():string {
  return (typeof process !== "undefined" ? process.env.YAGNA_API_URL : "") || "http://127.0.0.1:7465";
}
```

yajsapi/payment/config.ts
```diff
@@ -110,13 +108,13 @@ export class InvoiceConfig extends BaseConfig {
 }
 
 export class AccountConfig extends BaseConfig {
```
mgordel, line 110: If the tests pass and it is unnecessary, delete the comments

yajsapi/market/config.ts
```diff
@@ -31,13 +30,13 @@ export class DemandConfig {
...
-    const apiKey = options?.yagnaOptions?.apiKey || process.env.YAGNA_APPKEY;
+    const apiKey = options?.yagnaOptions?.apiKey || EnvUtils.getYagnaAppKey();
     if (!apiKey) throw new Error("Api key not defined");
-    const basePath = options?.yagnaOptions?.basePath || process.env.YAGNA_API_URL || DEFAULTS.basePath;
+    const basePath = options?.yagnaOptions?.basePath || EnvUtils.getYagnaApiUrl()
     const apiConfig = new Configuration({ apiKey, basePath: `${basePath}/market-api/v1`, accessToken: apiKey });
     this.yagnaOptions = options?.yagnaOptions;
     this.api = new RequestorApi(apiConfig);
-    this.subnetTag = options?.subnetTag || process.env.YAGNA_SUBNET || DEFAULTS.subnetTag;
+    this.subnetTag = options?.subnetTag || (typeof process !== 'undefined' ? process.env.YAGNA_SUBNET : '') || DEFAULTS.subnetTag;
```
mgordel, line 39: EnvUtils can be used here too

yajsapi/utils/env.ts
pgrzy-golem: it is really annoying mock plain functions, that is why I usually try to avoid those

Braunmann (approved):

PR #502, opened 2023-06-06
363 advanced web example

mgordel:


examples/web/requestor/assets/scss/element/index.scss
```diff
@@ -0,0 +1,33 @@
+//$-colors: (
```
pgrzy-golem, line 1: why is it not deleted?

pgrzy-golem (approved):

PR #505, opened 2023-06-12
Integrate with new registry

pociej:
- `TaskExecutor.create` can now be called with tag instead of hash string (hash is still supported). Library then checks based on string shape is it hash or tag. If tag then it calls registry server to provide hash for given tag and procedure continues as before with hash. 
-  `Package` can be constructed with `imageTag` instead of `imageHash`
-  repo url can be provided by `REPO_URL` env variable. For testing purpose please use [stage server](https://registry.dev.golem.network/)


.vscode/settings.json
mgordel, line 1: rather we don't want to store ide configs, it should go to .gitignore..?

tests/unit/package.test.ts
```diff
@@ -10,17 +10,21 @@ describe("Package", () => {
       expect(p).to.be.instanceof(Package);
     });
     it("should return decorators with task_package and package_format", async () => {
+      // Due to missing mocking and DI approach this tests is not mocked
+      // and makes real request to the registry
+      // ? Shouldnt we avoid this
```
mgordel, line 15: Yes, you're right, but for now, let it be.

yajsapi/events/events.ts
```diff
@@ -65,7 +66,14 @@ export class ProposalFailed extends BaseEvent<{
   reason?: string;
 }> {}
 export class ProposalConfirmed extends BaseEvent<{ id: string; providerId: string }> {}
-export class PackageCreated extends BaseEvent<{ imageHash?: string; details: PackageDetails }> {}
+export class PackageCreated extends BaseEvent<{
+  packageReference: RequireAtLeastOne<{
```
mgordel, line 70: Couldn't this be added to package details?

yajsapi/package/package.ts
```diff
@@ -59,10 +60,30 @@ export class Package {
   }
 
   static create(options: PackageOptions): Package {
+    // ? : Dependency Injection could be useful
```
mgordel, line 63: what do you think should be the dependencies?

yajsapi/package/package.ts
```diff
@@ -82,23 +103,36 @@ export class Package {
...
-    const response = await axios.get(`${repoUrl}/image.${this.options.imageHash}.link`);
+    const repoUrl = EnvUtils.getRepoUrl();
+
+    //TODO : in future this should be passed probably through config
+
+    const isHttps = process.env.YAJSAPI_USE_HTTPS_LINK ?? false;
+
+    // ? Should we prefix all env variables with YAJSAPI_ or not?
+    // with YAJSAPI we stay consistent but GOLEM is more general and can be used
+    // for other projects as well (yapapi e.g. )
```
mgordel, line 115: We will drop the name `YAJSAPI` so the `GOLEM` prefix is much better

yajsapi/package/package.ts
```diff
@@ -82,23 +103,36 @@ export class Package {
...
-  }
-
   private async resolveTaskPackageUrl(): Promise<string> {
-    const repoUrl = await this.getRepoUrl();
-    const response = await axios.get(`${repoUrl}/image.${this.options.imageHash}.link`);
+    const repoUrl = EnvUtils.getRepoUrl();
+
+    //TODO : in future this should be passed probably through config
+
+    const isHttps = process.env.YAJSAPI_USE_HTTPS_LINK ?? false;
```
mgordel, line 111: I think this should go to PackageOptions and default values to DEFAULTS

yajsapi/package/package.ts
```diff
@@ -82,23 +103,36 @@ export class Package {
...
+
+    // ? Should we prefix all env variables with YAJSAPI_ or not?
+    // with YAJSAPI we stay consistent but GOLEM is more general and can be used
+    // for other projects as well (yapapi e.g. )
+
+    const isDev = process.env.GOLEM_DEV_MODE;
```
mgordel, line 117: the same, it should rather be in the config as PackageOptions

yajsapi/package/repo_resolver.ts
mgordel, line 1: I think we can safely remove it

yajsapi/package/package.ts
mgordel, line 111: I think this should go to PackageOptions and default values to DEFAULTS

mgordel (approved):

PR #521, opened 2023-07-19
JST-22 GFTP over websockets

pgrzy-golem:


examples/web/Dockerfile.imagemagick
mgordel, line 1: In other examples we use e naming convention like `imagemagic.Dockerfile` or only `Dockerfile`

package.json
```diff
@@ -40,10 +40,13 @@
     "bottleneck": "^2.19.5",
     "collect.js": "^4.34.3",
     "eventsource": "^2.0.2",
+    "flatbuffers": "^23.5.26",
     "ip-num": "^1.4.1",
+    "js-sha3": "^0.8.0",
```
mgordel, line 45: I understand that we use an external library due to the low efficiency of the native hashing function from the nodejs crypto module?

yajsapi/storage/ws-browser.ts
```diff
@@ -0,0 +1,234 @@
...
+ */
+export class WebSocketBrowserStorageProvider implements StorageProvider {
+  /**
+   * Map of open services (IDs) indexed by GFTP url.
+   */
+  private services = new Map<string, string>();
+  private logger: Logger;
+
+  constructor(private readonly options: WebSocketStorageProviderOptions) {
+    this.logger = options.logger ?? nullLogger();
```
mgordel, line 68: Is it necessary to use null Logger? we haven't done that before and the logger was always nullable..

yajsapi/storage/ws-browser.ts
```diff
@@ -0,0 +1,234 @@
...
+      } else {
+        this.logger.warn(`[WebSocketBrowserStorageProvider] Unsupported message in publishData(): ${(req as GsbRequest<void>).component}`);
+      }
+    });
+
+    return fileInfo.url;
+  }
+
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
mgordel, line 101: I don't think we should disable tslint in this way, let's just remove this parameter. TypeScript will allow it

yajsapi/storage/ws-browser.ts
```diff
@@ -0,0 +1,234 @@
...
+        callback(result);
+      } else {
+        this.logger.warn(`[WebSocketBrowserStorageProvider] Unsupported message in receiveData(): ${(req as GsbRequest<void>).component}`);
+      }
+    });
+
+    return fileInfo.url;
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
mgordel, line 128: Same as above

examples/web/executor/image.html
mgordel, line 1: I think we should now do a cypress integration test for this example. Previously, it was not much different from HelloWorld. Now we should test it end2end.

mgordel (changes requested):

package.json
pgrzy-golem: No, this is in browser, SHA3 is not available there so we use this low efficiency module instead.

yajsapi/storage/ws-browser.ts
pgrzy-golem: Having null objects for complex object with logic make the code cleaner, you don't need to check for it's existence every time you use it and the code can simply assume it's always there.

Null logger was introduced in logger improvements, and similarly null storage provider was introduced in storage provider improvements (though probably not really a null object itself)

mgordel (approved):

PR #524, opened 2023-07-25
Executor.end exits requestor scripts (JST-75)

mgordel:


yajsapi/executor/executor.ts
```diff
@@ -376,35 +378,24 @@ export class TaskExecutor {
...
-    await this.end()
-      .then(() => {
-        if (this.options.isSubprocess) throw new Error(message);
-        else if (runtimeContextChecker.isNode) process.exit(0);
-      })
-      .catch((error) => {
-        this.logger?.error(error);
-        if (runtimeContextChecker.isNode && !this.options.isSubprocess) process.exit(1);
-      });
+    await this.end().catch((error) => this.logger?.error(error));
```
grisha87, line 398: Please don't mix `await` and `catch` like that - it's mixing coding styles and making things harder to read, understand and maintain. Instead consider:

```ts
  public async cancel(reason?: string) {
    try {
      if (this.isCanceled) return;
      if (runtimeContextChecker.isNode) this.removeCancelEvent();
      const message = `Executor has interrupted by the user. Reason: ${reason}.`;
      this.logger?.warn(`${message}. Stopping all tasks...`);
      this.isCanceled = true;
      await this.end();
    } catch (err) {
      this.logger?.error(err, "Error while cancelling the executor");
    }
  }
```

yajsapi/payment/service.ts
```diff
@@ -83,6 +83,7 @@ export class PaymentService {
     this.payments?.unsubscribe().catch((error) => this.logger?.warn(error));
     this.payments?.removeEventListener(PaymentEventType, this.subscribePayments.bind(this));
     for (const allocation of this.allocations) await allocation.release().catch((error) => this.logger?.warn(error));
+    this.options.httpAgent.destroy?.();
```
grisha87, line 86: Why do you need the `?` here? This method is always present if you built the HTTP agent correctly in the constructor. I checked this locally and the code compiles `this.options.httpAgent.destroy();` without any issues.

yajsapi/agreement/config.ts
```diff
@@ -21,16 +21,18 @@ export class AgreementConfig {
   readonly api: RequestorApi;
   readonly logger?: Logger;
   readonly eventTarget?: EventTarget;
+  public readonly httpAgent: Agent;
```
grisha87, line 24: Looks like `private` to me...

yajsapi/market/config.ts
```diff
@@ -31,16 +31,18 @@ export class DemandConfig {
   public readonly proposalTimeout?: number;
   public readonly logger?: Logger;
   public readonly eventTarget?: EventTarget;
+  public readonly httpAgent: Agent;
 
   constructor(options?: DemandOptions) {
     const apiKey = options?.yagnaOptions?.apiKey || EnvUtils.getYagnaAppKey();
     if (!apiKey) throw new Error("Api key not defined");
     const basePath = options?.yagnaOptions?.basePath || EnvUtils.getYagnaApiUrl();
+    this.httpAgent = new Agent({ keepAlive: true });
```
grisha87, line 40: It would be best if a single agent instance would be created and passed down via constructors of these objects. Right now we produce few of them, which means we need to maintain all instances and ensure that they get closed. Consider:

```ts
constructor(
  private readonly httpAgent: Agent
  options?: DemandOptions,
)
```

It will also make it more "mockable" later in tests.

yajsapi/network/config.ts
```diff
@@ -20,17 +20,19 @@ export class NetworkConfig {
...
   public readonly logger?: Logger;
   public readonly apiUrl: string;
+  public readonly httpAgent: Agent;
 
   constructor(options: NetworkOptions) {
     const apiKey = options?.yagnaOptions?.apiKey || EnvUtils.getYagnaAppKey();
     if (!apiKey) throw new Error("Api key not defined");
     const basePath = options?.yagnaOptions?.basePath || EnvUtils.getYagnaApiUrl();
     this.apiUrl = `${basePath}/net-api/v1`;
+    this.httpAgent = new Agent({ keepAlive: true });
```
grisha87, line 30: Similar note as in other cases, pass this via constructor parameters.

grisha87 (changes requested):

yajsapi/executor/executor.ts
mgordel: yes you may be right i will change it here but unfortunately we have it in many places...

yajsapi/payment/service.ts
mgordel: There is a polyfill in the browser and unfortunately it does not have this method:

https://github.com/FredKSchott/rollup-plugin-polyfill-node/blob/eb507a21ad82236255988548dea6120d7a277de2/polyfills/http.js#L71

We made httpAgent mainly because in nodejs there were random errors in connections to yagna - socket hangup and httpAgent resolve this problem. It wasn't in the browser.

yajsapi/agreement/config.ts
mgordel: I don't think so. We use this object later in entities, eg. Agreement, but maybe the name is misleading...

`this.options.httpAgent.destroy?.();`

yajsapi/market/config.ts
mgordel: Yes you are right. The only problem is that in the original idea we assumed that core entities such as Agreemnet, Activity, etc. can exist independently (mid-level API). With your approach we would have to create one object at the top (in Executor) and pass it down. In addition, destroy would also have to be done on top.. 
This will cause each entity and service to have an additional dependency on the constructor and maybe there would be nothing wrong with it, but I am not sure if it will be needed in future long pooling and the user (who would use services and maybe low-entities) should know what is httpAgent so for now I would leave it that way

mgordel: We should also consider making a global config object for connections to yagna to simplify mocking in tests. Each entity uses the same connection and creates it for itself every time.. So maybe with these fixes it can be improved

yajsapi/executor/executor.ts
grisha87: One step at a time.

grisha87:
Changes from this PR have been included to 0.7.1, and will be merged together with that branch.

PR #527, opened 2023-07-27
JST-112: Drop imports using .js extensions and…

grisha87:
adjust rollup config to produce separate artifacts for browser and nodejs runtimes


rollup.config.mjs
```diff
@@ -0,0 +1,51 @@
...
+          {find: /IORedisConnection/, replacement: "tests/mock/utils/empty_default.js"},
+          {find: /RedisConnection/, replacement: "tests/mock/utils/empty_default.js"},
+          {find: /eventsource\/lib\/eventsource.js/, replacement: "tests/mock/utils/empty_default.js"},
+          {find: /src\/api\/provider-api$/, replacement: "."},
+          {find: /\.\/gftp.js/, replacement: "tests/mock/utils/empty.js"},
+        ],
+      }),
+      nodeResolve({browser: true, preferBuiltins: true}),
+      commonjs(),
+      nodePolyfills(), // ToDo Needed?
```
grisha87, line 36: Deal with todos, @mgordel , can you provide me with more context for some of those? Why we had to introduce them? Can some of them be dropped?

grisha87:
I was debugging the issue with `ts-mocha` and imports in tests. It seems that this problem is there only because mocha, because the same test designed for jest works fine:

```ts
import {add} from "./check";

describe("Example test", () => {
  test("a+b", () => {
    expect(add(1, 2)).toEqual(3);
  })
});
```

While in case of ts-mocha, that `import` is illegal. Seems to be a known problem:
https://stackoverflow.com/questions/60993812/mocha-typescript-cannot-use-import-statement-outside-a-module

rollup.config.mjs
pgrzy-golem: long term I think we should drop stream support and move to native browser stuff with polyfills.

if some native node functionality is needed, we make a separate lib for it

.gitignore
```diff
@@ -29,4 +29,7 @@ package-lock.json
 .vscode
 
 #Nuxt
-/examples/adv-web-example/.nuxt
\ No newline at end of file
+/examples/adv-web-example/.nuxt
+
+#Goth
+tests/goth/assets/
```
grisha87, line 35: That's actually my own fault.

yajsapi/package/repo_resolver.ts
mgordel, line 1: This file is to be deleted, no longer needed

mgordel (approved):

PR #532, opened 2023-07-31
JST-63: Removed throwing error when results contain error

mgordel:


tests/integration/tasks.spec.ts
```diff
@@ -33,6 +33,21 @@ describe("Task Executor", function () {
...
+      payment: { network: "rinkeby" },
+      logger,
+    });
+    const result1 = await executor.run(async (ctx) => ctx.run("echo 'Hello World'"));
+    const result2 = await executor.run(async (ctx) => ctx.run("invalid-command"));
+
+    expect(result1?.stdout).to.include("Hello World");
+    expect(result2?.result).to.equal("Error");
+    expect(result2?.stderr).to.include("sh: 1: invalid-command: not found");
+    expect(result2?.message).to.equal("ExeScript command exited with code 127");
```
pgrzy-golem, line 48: you could also validate the status of the results

pgrzy-golem (changes requested):

tests/integration/tasks.spec.ts
mgordel: Which status do you mean? In the Result type we only have: `result`, `stdout`, `stderr` and `message`...

pgrzy-golem (approved):

PR #537, opened 2023-08-07
Typedocs task api snippets fixes and unhide reference options

mgordel:


yajsapi/executor/executor.ts
```diff
@@ -268,7 +265,7 @@ export class TaskExecutor {
    * @example
    * ```typescript
    * const data = [1, 2, 3, 4, 5];
-   * const results = executor.map(data, (ctx, item) => providerCtx.ctx(`echo "${item}"`));
+   * const results = executor.map(data, async (ctx, item) => ctx.run(`echo "${item}"`));
```
grisha87, line 268: There's no use in `async` if you don't `await` anything in the body. of the function.

grisha87 (approved):

PR #542, opened 2023-08-08
Update summary generator to support new docs generating

cryptobench:
The goal of this PR was to update the code so that it doesn't produce a hidden file named .SUMMARY.md , but instead it produces an overview.md file which is what will be used to display the contents of the reference on the new documentation software.

And further more you can specify the output directory. This was done because i have the docs generation script in the golem-docs repository which clones yajsapi to a temp directory and then generates everything and cleans it up.


.docs/summary-generator.cjs
```diff
@@ -18,7 +17,9 @@ function capitalizeFirstLetter(string) {
 }
 
 (async () => {
-  const logStream = fs.createWriteStream(path.join(directoryPath, ".SUMMARY.md"), { flags: "w" });
+  const docsDir = process.argv[2] || "docs";
+  const directoryPath = path.join(__dirname, "..", docsDir);
+  const logStream = fs.createWriteStream(path.join(process.argv[3], "overview.md"), { flags: "w" });
```
mgordel, line 22: I get an error: `The "path" argument must be of type string. Received undefined` if i run it withouts args: `npm run docs:md`. Because `process.argv[3]` is undefined and must be a string. here I think it should be: `path.join(process.argv[3] || '"", "overview.md")` or rather `path.join(directoryPath, "overview.md")` .. ?

.docs/summary-generator.cjs
```diff
@@ -1,14 +1,13 @@
 const path = require("path");
 const fs = require("fs");
-const docsDir = "docs";
-const directoryPath = path.join(__dirname, "..", docsDir);
 
-async function prepareDocAnchor(type, file) {
+async function prepareDocAnchor(docsDir, type, file) {
+  const directoryPath = path.join(docsDir);
   const filePath = path.join(directoryPath, type, file);
   const fileContent = await fs.readFileSync(filePath, "utf-8");
```
mgordel, line 7: 'await' is unnecesary here, but it was here before, but it can be improved by the way..

.docs/summary-generator.cjs
```diff
@@ -1,14 +1,13 @@
 const path = require("path");
 const fs = require("fs");
-const docsDir = "docs";
-const directoryPath = path.join(__dirname, "..", docsDir);
 
-async function prepareDocAnchor(type, file) {
+async function prepareDocAnchor(docsDir, type, file) {
+  const directoryPath = path.join(docsDir);
```
mgordel, line 5: this is probably redundant. It could be: `const filePath = path.join(docsDir, type, file);`

mgordel (changes requested):

mgordel (approved):

PR #550, opened 2023-08-11
JST-70: Automate releases

grisha87:
This PR introduces `semantic-release` to the project.

As part of this PR we:

- separated a fast CI pipeline which is concerned about formatting, linting and unit tests
- separated a slow E2E and release pipeline which is there to ensure more testing before the automated release
- dropped tests on Node 19 as it's not supported anymore
- this PR requires from developers to follow the conventional commit message formatting
- Goth and Cypress tests will be runnable manually (triggering from the UI)

Thanks to these changes we gain:

- shorter feedback loops for PRs
- automated releases to GH, NPM including automated generation of release notes

Closes #173


.github/workflows/ci.yml
```diff
@@ -0,0 +1,41 @@
...
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: Perform regular checks
+        run: |
+          npm install
+          npm run format:check
+          npm run lint
+          npm run test:unit
+          npm run build
```
mgordel, line 41: Maybe we should do a build before units first? if TSC doesn't pass, there's no need to run tests..

.github/workflows/release.yml
```diff
@@ -0,0 +1,197 @@
...
+on:
+  push:
+    branches:
+      # Regular release channels
+      - master
+      - next
+      - beta
+      - alpha
+      # Support, hotfix branches like: 1.0.x or 1.x
+      - "+([0-9])?(.{+([0-9]),x}).x"
```
mgordel, line 12: is this regex correct?
I'm not sure but I tested on validator and I think it should be: `([0-9])+(.{+([0-9]),x})?.x`

.github/workflows/release.yml
```diff
@@ -0,0 +1,197 @@
...
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: Perform regular checks
+        run: |
+          npm install
+          npm run format:check
+          npm run lint
+          npm run test:unit
```
mgordel, line 44: as above .. ?

mgordel (approved):

.github/workflows/ci.yml
grisha87: As per Discord -> test:unit will force tsc due to `ts-jest`, so transpilation errors will fail the build early.

.github/workflows/release.yml
grisha87: Nice catch! That's actually a pattern for `micromatch`, so I patched the regex used in the workflow files to a valid reg-exp.

.github/workflows/ci.yml
mgordel: You're right

PR #560, opened 2023-08-23
tests e2e timeout

mgordel:


grisha87 (approved):

grisha87:
@mgordel , I switched the target to master, tested by triggering the workflow and it works fine now. Going to merge it to master first, and then back to beta.

PR #579, opened 2023-09-07
fix(docs): fix summary generator not creating a name

cryptobench:
The latest changes in the frontmatter generation plugin for the docs broke the summary generator. All markdown links generated now has no names, but the links are still working.

This PR solves that issue and adds frontmatter to the content overview file so that the page doesn't have a title named `undefined`


grisha87 (changes requested):
Please reword the commit: `fix(docs): fix summary generator not creating a name` into `docs: fix summary generator not creating a name` so that we don't generate an unnecessary release that doesn't really fix anything in the SDK.

PR #584, opened 2023-09-11
JST-398 not all config is passed to executor

SewerynKras:


src/golem_network/golem_network.ts
```diff
@@ -1,12 +1,38 @@
-import { TaskExecutor } from "../executor";
+import { TaskExecutor, YagnaOptions } from "../executor";
+import { JobStorage } from "../job";
 import { PackageOptions } from "../package";
 import { Worker } from "../task";
 
 export interface GolemNetworkConfig {
+  /**
+   * Image that will be uploaded to the provider and used to run the task. Defaults to `golem/alpine:3.18.2`.
```
mgordel, line 8: I think it should be `golem/alpine:latest`

src/golem_network/golem_network.ts
```diff
@@ -53,12 +70,14 @@ export class GolemNetwork {
 
   public async init() {
     this._executor = await TaskExecutor.create({
-      package: this.image,
-      enableLogging: this.enableLogging,
-      ...this.demand,
+      package: this.config.image ?? "golem/alpine:3.18.2",
```
mgordel, line 73: `golem/alpine:latest` ?

src/golem_network/golem_network.ts
```diff
@@ -1,12 +1,38 @@
-import { TaskExecutor } from "../executor";
+import { TaskExecutor, YagnaOptions } from "../executor";
+import { JobStorage } from "../job";
 import { PackageOptions } from "../package";
 import { Worker } from "../task";
 
 export interface GolemNetworkConfig {
```
mgordel, line 6: maybe it's not for this time, but in all old entities we use the `xxxOptions` convention to name interfaces, so maybe it would be better to use `GolemNetworkOptions` for consistency..?

mgordel (approved):

PR #589, opened 2023-09-13
e2e test fixes

mgordel:
Running goth tests for the new hascat image `golem/examples-hashcat:latest` causes errors on goth providers: `* Device #1: Not enough allocatable device memory for this attack.`
Therefore, we have to leave the old image hash for hashcat test.


tests/e2e/yacat.spec.ts
```diff
@@ -17,7 +17,7 @@ describe("Password cracking", function () {
     const mask = "?a?a";
     const hash = "$P$5ZDzPE45CigTC6EY4cXbyJSLj/pGee0";
     executor = await TaskExecutor.create({
-      package: "golem/examples-hashcat:legacy",
+      package: "055911c811e56da4d75ffc928361a78ed13077933ffa8320fb1ec2db",
```
SewerynKras, line 20: Probably worth leaving a comment here for future generations why this test must use a specific hash rather than the latest image

SewerynKras (approved):

.github/workflows/goth-nightly.yml
```diff
@@ -34,7 +34,7 @@ jobs:
       - name: Configure node.js
         uses: actions/setup-node@v3
         with:
-          node-version: 16
```
pgrzy-golem, line 37: Why are we changing version?

mgordel: node16 is no longer supported. We are planning to change versions to 18 everywhere and when I touched the pipeline file I thought I would change it.

PR #607, opened 2023-10-04
fix(executor): added error when no proposals are received

mgordel:
JST-462: Added startupTimeout (default 30sec) as TaskExecutor option, after which if there is no at least one proposal we will get an appropriate error


src/executor/executor.ts
```diff
@@ -54,6 +54,8 @@ export type ExecutorOptions = {
    * For more details see {@link JobStorage}. Defaults to a simple in-memory storage.
    */
   jobStorage?: JobStorage;
+  /** Timeout for waiting for at least one offer from the market  */
```
grisha87, line 57: Please provide a more user friendly description  in the second line, explaining why the user should care and what they can do with this setting, what will happen when the timeout is reached. Documenting it here is far more important than writing the article in the docs :)

src/executor/executor.ts
```diff
@@ -385,11 +390,29 @@ export class TaskExecutor {
       timeout: options?.timeout ?? this.options.taskTimeout,
     });
     this.taskQueue.addToEnd(task as Task<unknown, unknown>);
+    if (!this.startupTimeoutError && !this.startupTimeoutId) {
```
grisha87, line 393: I have few issues with this, we can talk about this f2f:

- This way of communicating the error with the timeout isn't easy for the user to handle, we should work on that
- Since this is implemented as part of `executeTask`, this timeout will be started and counted for each scheduled task. I do see a concurrency problem here, given all the tasks are executed on the same TaskExecutor instance.

src/market/demand.ts
```diff
@@ -27,6 +27,7 @@ export interface DemandOptions {
   offerFetchingInterval?: number;
   proposalTimeout?: number;
   eventTarget?: EventTarget;
+  startupTimeout?: number;
```
grisha87, line 30: This field should be documented as well.

src/market/service.ts
```diff
@@ -28,6 +27,12 @@ export class MarketService {
   private logger?: Logger;
   private taskPackage?: Package;
   private maxResubscribeRetries = 5;
+  private proposalsCount = {
+    initial: 0,
+    filtered: 0,
```
grisha87, line 32: Filtered means accepted or rejected? The naming is confusing.

grisha87 (changes requested):

src/executor/executor.ts
mgordel: you're right, I'll add a better description

src/market/demand.ts
mgordel: this is unnecessary, left by mistake, i will delete it

src/market/service.ts
mgordel: filtered is not used, it was left by mistake

src/executor/executor.ts
mgordel: this timeout will only start once, unless `this.startupTimeoutId` exists yet, so there shouldn't be a problem with concurrency, I think, but let's talk about it..

grisha87 (changes requested):
As discussed during the call, please move the timeout tracking logic out of the "task execution phase", back to the initialisation.

grisha87 (approved):

PR #610, opened 2023-10-04
Feature/jst 468/make use of public whitelist

grisha87:


mgordel (approved):
The only problem may be returning an empty array and passing it to the whitelist filter. If any error occurs while fetching the providers, we return an empty array, which will block the entire executor. But I think it is described so well in the comments that the user in such critical situations can handle it himself..
And also this `console.error` which will not be in the streams defined in the logger configuration, but I don't see a good solution if we want to keep it as a simple helper (without dependencies)

so ok for me

src/market/helpers.test.ts
```diff
@@ -30,13 +30,13 @@ describe("Market Helpers", () => {
...
-
-        // Then
-        expect(data).toEqual([]);
+        // When, Then
+        await expect(() => getHealthyProvidersWhiteList()).rejects.toThrow(
+          "Failed to download healthy provider whitelist due to an error: Error: Request to download healthy provider whitelist failed: {error:'test'}",
+        );
       });
 
       test("If the implementation will throw any kind of error, then it will return an empty array", async () => {
```
mgordel, line 42: this test description is no longer valid

mgordel (approved):
yes, now the user has to handle the situation where something goes wrong and decide for himself. For me this solution is better.

src/market/helpers.test.ts
grisha87: Thanks for pointing this out! I just updated the test names :)

PR #612, opened 2023-10-05
JST-429: Gftp bugfix and refactor

mgordel:
JST-429: Fixed gftp bug with incorrect initialization when there are no tasks in the executor.
Additionally, gftp has been refactored, getting rid of the old `@rauschma/stringio` dependency which has been replaced with native nodejs modules.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
     name: Build and unit-test on supported platforms and NodeJS versions
     strategy:
       matrix:
-        node-version: [16.x, 18.x, 20.x]
+        node-version: [18.x, 20.x]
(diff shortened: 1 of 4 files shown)


pgrzy-golem (approved):
One thing I'm concerned about is the use of ? in declaration and later usage. 
What this does is that it will silently ignore a lot of issues at runtime.

We should have a discussion about our coding style, and consider either using null values/default objects, or `!` instead or choosing a completely different approach.

mgordel:
@pgrzy-golem I agree that it's not a very nice style, but in a situation where we initialize the object later after its creation, this will be a problem and will require refactoring of objects, so we can discuss it and agree on a common approach, but maybe as part of another task

PR #628, opened 2023-10-23
Change default driver to erc20next

kamirr:
Only to be merged after payments UX release.


grisha87:
@kamirr , which version of Yagna is going to support this?

grisha87:
As discussed with @prekucki and @nieznanysprawiciel last week - the driver change will be transparent for the SDKs and solved within yagna itself. Thus, this rename in the SDK is not required anymore.

PR #635, opened 2023-10-30
Streaming of command results

mgordel:


src/task/process.ts
```diff
@@ -0,0 +1,65 @@
...
+    this.streamOfActivityResults.on("data", (data) => (this.lastResult = data));
+    this.streamOfActivityResults.on("error", (error) => (this.streamError = error));
+    const { stdout, stderr } = this.transformResultsStream();
+    this.stdout = stdout;
+    this.stderr = stderr;
+  }
+
+  /**
+   * Waits for the process to complete and returns the last part of the command's results as a {@link Result} object
+   * @param timeout - maximum waiting time for the final result (default: 20 sec)
```
pgrzy-golem, line 31: Please add something about the error, like: If timeout is reaached, returned promise will be rejected.

src/task/process.ts
```diff
@@ -0,0 +1,65 @@
...
+   */
+  waitForExit(timeout?: number): Promise<Result> {
+    return new Promise((res, rej) => {
```
pgrzy-golem, line 34: can we use full names for resolve and reject? I think it gives a more meaningful name for the variables

src/task/process.ts
```diff
@@ -0,0 +1,65 @@
...
+      const timeoutId = setTimeout(
+        () => rej(new Error("The waiting time for the final result has been exceeded")),
```
pgrzy-golem, line 36: I would include the timeout value in the message, so if user didn't his own timeout, he'll know what he is working with

src/task/process.ts
```diff
@@ -0,0 +1,65 @@
...
+    const { stdout, stderr } = this.transformResultsStream();
+    this.stdout = stdout;
+    this.stderr = stderr;
+  }
+
+  /**
+   * Waits for the process to complete and returns the last part of the command's results as a {@link Result} object
+   * @param timeout - maximum waiting time for the final result (default: 20 sec)
+   */
+  waitForExit(timeout?: number): Promise<Result> {
```
pgrzy-golem, line 33: I'm worried that this method produces two distinct errors:
1. Timeout error
2. Error if lastResult is missing
I think we should be able to tell the difference, at least in this specific case.
We should discuss this topic on some refinement maybe?

tests/e2e/tasks.spec.ts
```diff
@@ -180,4 +179,28 @@ describe("Task Executor", function () {
     const result = await executor.run(async (ctx) => ctx.getIp());
     expect(["192.168.0.2", "192.168.0.3"]).toContain(result);
   });
+
+  it("should run simple by spawn command as external process", async () => {
```
pgrzy-golem, line 183: "run simple by spawn" ?

pgrzy-golem (changes requested):

src/task/process.ts
mgordel: I agree, but we need to define different Error types throughout the project

pgrzy-golem (approved):

PR #637, opened 2023-11-02
JST-563 Jobs that run without a TaskExecutor underneath

SewerynKras:


examples/express/server.ts
```diff
@@ -1,127 +1,114 @@
...
+    },
+  });
+
+  job.events.on("created", () => {
+    console.log("Job created");
+  });
+  job.events.on("started", () => {
+    console.log("Job started");
+  });
+  job.events.on("error", () => {
```
mgordel, line 42: maybe it would be natural to get an error in the parameter `job.events.on("error", (error) => {`

src/golem_network/golem_network.ts
```diff
@@ -1,125 +1,67 @@
...
-    });
-    if (this.config.beforeEachJob) {
-      this.executor.onActivityReady(this.config.beforeEachJob);
+    if (this._yagna !== null) {
+      return;
     }
+    const yagna = new Yagna(this.config.yagna);
+    // make sure it's possible to connect to the yagna service
+    // this will throw an error if yagna is not running
+    await yagna.getApi().identity.getIdentity();
```
mgordel, line 38: There is a function `yagna.connect()` that does the same thing

src/job/job.ts
```diff
@@ -1,43 +1,237 @@
...
+
+  /**
+   * Cancel the job. This method will stop the activity and wait for it to finish.
+   * Throws an error if the job is not running.
+   */
+  async cancel() {
+    if (!this.isRunning || !this.activity) {
+      throw new Error(`Job ${this.id} is not running`);
+    }
+    await this.activity.stop();
```
mgordel, line 211: Here we should also release the agreement: `await agreementPoolService.releaseAgreement(...)`

mgordel (approved):
`Job` class is very similar to my [GolemRuntime](https://github.com/golemfactory/golem-sdk-worker/blob/master/src/runtime.ts) class in [golem-sdk-worker](https://github.com/golemfactory/golem-sdk-worker/tree/master).. maybe in the future it will be possible to separate it into one common calss..

Overall, I don't see any critical errors. Just a few minor comments...

SewerynKras:
Turns out, just calling `activity.stop` does not actually cancel the work if it hasn't fully started yet, so I implemented a proper cancellation mechanism with an `AbortController` and wrote a unit test to verify. 
I also added a few more examples

PR #643, opened 2023-11-08
Throw an error when a task fails instead of ending the executor

SewerynKras:


src/executor/executor.spec.ts
```diff
@@ -93,6 +93,47 @@ describe("Task Executor", () => {
...
+      const errorSpy = jest.spyOn(Task.prototype, "getError");
+
+      rejectedSpy.mockImplementationOnce(() => false);
+      resultsSpy.mockImplementationOnce(() => "result 1");
+      const result1 = await executor.run(() => Promise.resolve());
+
+      try {
+        rejectedSpy.mockImplementationOnce(() => true);
+        errorSpy.mockImplementationOnce(() => new Error("test"));
+        await executor.run(() => Promise.resolve());
```
grisha87, line 115: This can be replaced with something like `await expect(executor.run(() => throw new Error("test")).rejects.toEqual("test")`, then you can get rid of the whole `try/catch` block here. Then the `onError` spy also becomes unnecessary.

grisha87 (approved):

src/executor/executor.spec.ts
SewerynKras: Good idea. I also changed the other calls to `executor.run` to use `await expect(...)` for consistency

grisha87 (approved):

PR #649, opened 2023-11-10
Improvements implemented while building Golem Worker Runtime

mgordel:
Improvements needed to properly implement `GolemRuntime` in repo [golem-sdk-worker](https://github.com/golemfactory/golem-sdk-worker).

The main changes are the replacement of dependencies in `Activity`. Instead of `agreemnetId`, the entire `Agreement` entity is now passed.
Additionally, a simple ActivityPool service has been implemented, which allows you to return an existing activity from the pool or create a new one if the pool is empty, and return it to the pool after finishing the work.
In the future, it will be possible to use this service in `TaskExecutor` or when refactoring `WorkContext`


src/activity/service.ts
```diff
@@ -0,0 +1,52 @@
+import { Activity, ActivityOptions } from "./activity";
+import { defaultLogger, Logger, YagnaApi } from "../utils";
+import { AgreementPoolService } from "../agreement";
+import { PaymentService } from "../payment";
+
+interface ActivityServiceOptions extends ActivityOptions {}
+
+export class ActivityPoolService {
```
grisha87, line 8: Looks like the file name has to be adjusted.

src/package/package.ts
```diff
@@ -4,35 +4,34 @@ import axios from "axios";
...
-    manifestSigAlgorithm?: string;
-    /** Certificate - base64 encoded public certificate (DER or PEM) matching key used to generate signature **/
-    manifestCert?: string;
-    logger?: Logger;
-  },
-  "imageHash" | "imageTag" | "manifest"
->;
+export type AllPackageOptions = {
+  /** Type of engine required: vm, emscripten, sgx, sgx-js, sgx-wasm, sgx-wasi */
+  engine?: string;
```
grisha87, line 9: Be specific instead of string then?

tests/mock/fixtures/worker.js
```diff
@@ -0,0 +1,13 @@
+import { parentPort } from "worker_threads";
```
grisha87, line 1: This file should be also moved to the other project

tests/mock/utils/empty.js
```diff
@@ -1,3 +1,4 @@
 export const GftpStorageProvider = {};
 export const pinoLogger = {};
 export const EventSource = {};
+export const GolemWorkerNode = {};
```
grisha87, line 4: To be moved away from this project

tests/unit/activity.test.ts
```diff
@@ -14,7 +15,7 @@ describe("Activity", () => {
 
   describe("Creating", () => {
     it("should create activity", async () => {
-      const activity = await Activity.create("test_agreement_id", yagnaApi);
+      const activity = await Activity.create(agreement, yagnaApi);
```
grisha87, line 18: The fact that `yagnaApi` has to be passed here all the time is a bit disturbing.

grisha87 (changes requested):

.eslintrc
```diff
@@ -2,7 +2,9 @@
   "root": true,
   "parser": "@typescript-eslint/parser",
   "env": {
-    "node": true
+    "node": true,
+    "worker": true,
```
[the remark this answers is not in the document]
grisha87: Adjusted this to be `shared-node-browser`

grisha87 (changes requested):
I've reviewed this code and separated SDK relevant changes from the implementation of the Golem Worker runtime. The Golem Worker runtime is going to be placed in the other repository: https://github.com/golemfactory/golem-sdk-worker . In this PR we only focus on generic improvements to the `golem-js` library.

In addition, after all of the changes are incorporated, this PR should be squashed in to only a handful of commits.

src/activity/service.ts
mgordel: This is very analogous to `AgreementPoolService`. This is a very primitive implementation of pool. It has a `get` and `release` method. After the task is finished, the activity is returned to the pool for reuse or terminated (if allowReuse = false)
We can change the name to `ActivityService`, but I think it implements a very simple pooling mechanism. In its current form, this service is only used by `GolemRuntime` in `golem-sdk-worker`, but it can be used (and extended) in `TaskExecutor` or `WorkContext` in the future.

src/activity/service.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    activityService = new ActivityPoolService(yagnaApi, agreementPoolServiceMock, paymentServiceMock, {
+      logger,
+    });
+    logger.clear();
+  });
+
+  describe("run()", () => {
+    it("should start service", async () => {
+      await activityService.run();
+      expect(logger.logs).toContain("Activity Pool Service has started");
```
SewerynKras, line 19: I don't like that we're testing the logger here, why not check if `isServiceRunning` is true instead?

src/activity/service.test.ts
```diff
@@ -0,0 +1,64 @@
...
+      await activityService.end();
+    });
+  });
+  describe("end()", () => {
+    it("should stop service", async () => {
+      await activityService.run();
+      await activityService.end();
+      expect(logger.logs).toContain("Activity Pool Service has been stopped");
```
SewerynKras, line 27: Same here, testing the logger doesn't really tell us if the business logic inside works as expected. I would check here if every activity in the pool is cancelled and what happens if any of the calls to `cancel` reject?

src/activity/service.ts
```diff
@@ -0,0 +1,76 @@
...
+      this.logger.debug(`Activity ${activity.id} has been released and will be terminated`);
+    }
+  }
+
+  /**
+   * Stop the service and terminate all activities from the pool
+   */
+  async end() {
+    this.isServiceRunning = false;
+    this.pool.forEach((activity) => activity.stop().catch((e) => this.logger.warn(e)));
```
SewerynKras, line 67: why not wait for all activities to stop? IMO if someone calls `await activityPoolService.end()` they expect all activities to be fully stopped and cleaned up by the time that promise resolves

src/activity/service.ts
```diff
@@ -0,0 +1,76 @@
...
+    return this.pool.shift() || (await this.createActivity());
+  }
+
+  /**
+   * Release the activity back into the pool or if it is not reusable
+   * it will be terminated and the agreement will be released
+   * @param activity
+   * @param allowReuse
+   */
+  async releaseActivity(activity: Activity, allowReuse: boolean = true) {
```
SewerynKras, line 51: I think `activityPoolService.releaseActivity(activity, true)` is confusing to the reader. It should be obvious what each parameter does and just `true` doesn't tell me anything. IMO the last argument should be refactored to take an object `activityPoolService.releaseActivity(activity, { allowReuse: true})`

SewerynKras: also if it's always reused when `true` then maybe `resuse` is a better name than `allowReuse`

src/activity/service.test.ts
```diff
@@ -0,0 +1,64 @@
...
+      expect(activity1).toEqual(activity2);
+      await activityService.end();
+    });
+  });
+  describe("releaseActivity()", () => {
+    it("should return activity to the pool if allowReuse flag is true", async () => {
+      await activityService.run();
+      const activity = await activityService.getActivity();
+      await activityService.releaseActivity(activity, true);
+      expect(logger.logs).toContain(`Activity ${activity.id} has been released for reuse`);
```
SewerynKras, line 51: Same here, maybe check if the activity is in `activityService['pool']`?

src/activity/service.ts
```diff
@@ -0,0 +1,76 @@
...
+    if (allowReuse) {
+      this.pool.push(activity);
+      this.logger.debug(`Activity ${activity.id} has been released for reuse`);
+    } else {
+      await activity.stop().catch((e) => this.logger.warn(e));
+      await this.agreementService.releaseAgreement(activity.agreement.id, allowReuse);
```
SewerynKras, line 57: `allowReuse` is always `false` here?

src/activity/service.test.ts
```diff
@@ -0,0 +1,64 @@
...
+      await activityService.end();
+    });
+
+    it("should terminate activity if allowReuse flag is false", async () => {
+      await activityService.run();
+      const activity = await activityService.getActivity();
+      await activityService.releaseActivity(activity, false);
+      expect(logger.logs).toContain(`Activity ${activity.id} has been released and will be terminated`);
+      expect(logger.logs).toContain(`Activity ${activity.id} destroyed`);
```
SewerynKras, line 60: And here, why not check if `agreementService.releaseAgreement` has been called with the activity id?

tests/unit/work.test.ts
```diff
@@ -155,8 +156,8 @@ describe("Work Context", () => {
     });
 
     it("should catch error while executing batch as stream with invalid command", async () => {
-      const activity = await Activity.create("test_agreement_id", yagnaApi);
-      const worker = async (ctx: WorkContext) => ctx.beginBatch().run("invalid_shell_command").endStream();
+      const activity = await Activity.create(agreement, yagnaApi);
+      const worker = async (ctx) => ctx.beginBatch().run("invalid_shell_command").endStream();
```
SewerynKras, line 160: why remove `: WorkContext` from all workers in this file? If we make changes to `WorkContext` TS will not pick up that these files break

SewerynKras (changes requested):

src/activity/service.test.ts
mgordel: you are right, but this entire Service is actually a clone of AgreementServicePool, so I duplicated the same structore and tests (and code smells) that are there. but actually in the new file I can do it better now..

tests/unit/work.test.ts
mgordel: it was probably a mistake when merging the beta

SewerynKras (approved):
LGTM, just make sure to run tests again after resolving the merge conflict :)

grisha87 (approved):

PR #654, opened 2023-11-13
JST-548 | JST-551: deprecate foreach and map

SewerynKras:


src/executor/executor.ts
```diff
@@ -387,6 +399,19 @@ export class TaskExecutor {
   }
 
   /**
+   * @deprecated This method is marked for removal in a future release.
+   * Migrate your code by using `Array.map` and `Promise.all` instead.
```
grisha87, line 403: I think the "Migrate" part has to also be on the same line to get properly display in IDE - minor serverity,

grisha87 (approved):

src/executor/executor.ts
SewerynKras: In VSCode you always get the same message `'X' is deprecated` so here it makes no difference. Not sure about other IDEs

image

PR #657, opened 2023-11-14
feat: task executor emits lifecycle events through an EventEmitter

pgrzy-golem:
JST-526

### DIFF
--- a/examples/simple-usage/lifecycle.ts
+++ b/examples/simple-usage/lifecycle.ts
@@ -1,0 +1,23 @@
+import { TaskExecutor } from "@golem-sdk/golem-js";
+
+/**
+ * This example shows how to use the TaskExecutor lifecycle events.
+ */
+(async function main() {
+  console.log("Creating TaskExecutor and registering for lifecycle events...");
+  const executor = await TaskExecutor.create("golem/alpine:latest");
+
+  executor.events.on("terminating", () => {
+    console.log("Lifecycle: TaskExecutor is about to be terminated");
+  });
+
+  executor.events.on("terminated", () => {
+    console.log("Lifecycle: TaskExecutor is terminated. Exiting...");
+    process.exit(0);
+  });
+
+  console.log("Calling TaskExecutor.end()");
+  await executor.end();
+  // process.exit() is called in an event handler above, so this message should not be displayed.
+  console.log("Should not display this message.");
+})();
--- a/src/executor/events.ts
+++ b/src/executor/events.ts
@@ -1,0 +1,19 @@
+/**
+ * This interface describes events emitted by `TaskExecutor` through `TaskExecutor.events` object.
+ */
+export interface TaskExecutorEventsDict {
+  /**
+   * Fires when task executor is initialized and ready to be used.
+   */
+  initialized: () => void;
+
+  /**
+   * Fires when task executor is about to shut down, immediately after TaskExecutor.end() is called.
+   */
+  terminating: () => void;
+
+  /**
+   * Fires when task executor is completely terminated.
+   */
+  terminated: () => void;
+}
--- a/src/executor/executor.spec.ts
+++ b/src/executor/executor.spec.ts
@@ -91,6 +91,8 @@ describe("Task Executor", () => {
       );
       await executor.end();
     });
+
+    // TODO: test the "initialized" event, once init() is public and create() is deprecated.
   });
 
   describe("run()", () => {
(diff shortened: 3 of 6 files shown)


examples/simple-usage/lifecycle.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  executor.events.on("terminated", () => {
+    console.log("Lifecycle: TaskExecutor is terminated. Exiting...");
+    process.exit(0);
+  });
+
+  console.log("Calling TaskExecutor.end()");
+  await executor.end();
+  // process.exit() is called in an event handler above, so this message should not be displayed.
+  console.log("Should not display this message.");
```
SewerynKras, line 22: Pretty cool example with `process.exit` in the callback, but IMO we should not promote such practices

src/executor/executor.ts
```diff
@@ -101,6 +103,12 @@ export type YagnaOptions = {
  * A high-level module for defining and executing tasks in the golem network
  */
 export class TaskExecutor {
+  /**
+   * EventEmitter2 instance emitting TaskExecutor events.
```
SewerynKras, line 107: EventEmitter3*

examples/simple-usage/lifecycle.ts
pgrzy-golem: Actually, why not?

SewerynKras: In my opinion, using `process.exit` is in general a bad practice (because it prevents node from cleaning up async/ IO operations) and I avoid it unless I have a very good reason to use it. 

Also this particular example - write to stdout and exit immediately - is exactly what node docs advise *not* to do: https://nodejs.org/api/process.html#processexitcode

examples/simple-usage/lifecycle.ts
```diff
@@ -0,0 +1,23 @@
+import { TaskExecutor } from "@golem-sdk/golem-js";
+
+/**
+ * This example shows how to use the TaskExecutor lifecycle events.
+ */
+(async function main() {
+  console.log("Creating TaskExecutor and registering for lifecycle events...");
+  const executor = await TaskExecutor.create("golem/alpine:latest");
+
+  executor.events.on("terminating", () => {
```
grisha87, line 10: I don't think this event adds any value except for showcasing the mechanism. Developers are used to events which represent something that something "happened", not "is happening, or is about to happen". Please remove the `terminating` event.

examples/simple-usage/lifecycle.ts
grisha87: I support this argument. `process.exit` is frequently miss-used by developers and the hinted usage here is not a good one. In general, I think that this example is really serving a purpose of a unit test. You have a good unit test here already so there's no need to create that example at all.

src/executor/executor.spec.ts
```diff
@@ -91,6 +91,8 @@ describe("Task Executor", () => {
       );
       await executor.end();
     });
+
+    // TODO: test the "initialized" event, once init() is public and create() is deprecated.
```
grisha87, line 95: So perhaps mark it as `it.todo(...)`?

src/executor/executor.spec.ts
```diff
@@ -145,5 +147,31 @@ describe("Task Executor", () => {
...
+      let terminated = false;
+      let terminating = false;
+
+      executor.events.on("terminating", () => {
+        expect(terminating).toBe(false);
+        expect(terminated).toBe(false);
+        terminating = true;
+      });
+
+      executor.events.on("terminated", () => {
```
grisha87, line 162: Instead of `let` vars you could go with something more elegant:

```ts
const executor = await TaskExecutor.create({ package: "test", startupTimeout: 0, logger, yagnaOptions });

const onTerminated = jest.fn();

executor.events.on("terminated", onTerminated);

await executor.end();
await executor.end();

expect(onTerminated).toHaveBeenCalledTimes(1);
```

Also, the test name does not correspond to the contents and implementation.

src/executor/executor.ts
```diff
@@ -279,6 +288,7 @@ export class TaskExecutor {
    * @private
    */
   private async doEnd() {
+    this.events.emit("terminating");
```
grisha87, line 291: As noted previously, please remove this event.

grisha87 (changes requested):

examples/simple-usage/lifecycle.ts
grisha87, line 10: If we want to notify the user about a condition in which they should shut-down their application, we need to have a "industry standard approach implemented".

grisha87:
Counter-proposal :)

```ts
const te = new TaskExecutor({});

await te.init();
// emit -> ready

await te.quit();
// emit -> close
// emit -> end
```

src/executor/events.ts
```diff
@@ -0,0 +1,20 @@
...
+  /**
+   * Fires when task executor is initialized and ready to be used.
+   */
+  ready: () => void;
+
+  /**
+   * Fires when task executor is about to shut down, immediately after TaskExecutor.shutdown() is called.
+   *
+   */
+  beforeend: () => void;
```
grisha87, line 14: Please change it to `beforeEnd`

src/executor/events.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  /**
+   * Fires when task executor is completely terminated.
```
grisha87, line 17: From the documenting perspective, you could add just a bit more information like: explaining that the pending invoices are checked and paid before this happens etc. Instead of documenting this in the examples and docs, it would be great to have this documented in the API reference.

src/executor/executor.ts
```diff
@@ -101,6 +103,12 @@ export type YagnaOptions = {
  * A high-level module for defining and executing tasks in the golem network
  */
 export class TaskExecutor {
+  /**
+   * EventEmitter (EventEmitter3) instance emitting TaskExecutor events.
+   * @see TaskExecutorEventsDict for available events.
+   */
+  readonly events: EventEmitter<TaskExecutorEventsDict> = new EventEmitter();
```
grisha87, line 110: I'm not sure if that was hinted in any of the commits: this kind of a change is a `BREAKING CHANGE` so please make sure that you use that statement right so that `semantic-release` can pick this up correctly.

grisha87 (approved):

PR #660, opened 2023-11-15
fix: fixed issue with too long result polling interval leading to batch timeout

grisha87:
Related issue: JST-357

### DIFF
--- a/src/script/script.ts
+++ b/src/script/script.ts
@@ -6,14 +6,16 @@ import { Result } from "../activity";
  * @hidden
  */
 export class Script {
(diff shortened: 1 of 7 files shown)


src/script/script.ts
```diff
@@ -6,14 +6,16 @@ import { Result } from "../activity";
  * @hidden
  */
 export class Script {
+  constructor(private commands: Command[] = []) {}
```
pgrzy-golem, line 9: I think the convention is to have statics above any instance method

pgrzy-golem (approved):

PR #673, opened 2023-11-21
JST-598: simplify worker type

SewerynKras:
BREAKING CHANGE: worker type no longer takes two parameters

### DIFF
--- a/src/executor/executor.ts
+++ b/src/executor/executor.ts
@@ -117,7 +117,7 @@ export class TaskExecutor {
   private networkService?: NetworkService;
   private statsService: StatsService;
   private activityReadySetupFunctions: Worker[] = [];
-  private taskQueue: TaskQueue<Task<unknown, unknown>>;
+  private taskQueue: TaskQueue<Task>;
   private storageProvider?: StorageProvider;
   private logger?: Logger;
   private lastTaskIndex = 0;
@@ -188,7 +188,7 @@ export class TaskExecutor {
     this.logger = this.options.logger;
     this.yagna = new Yagna(this.configOptions.yagnaOptions);
     const yagnaApi = this.yagna.getApi();
-    this.taskQueue = new TaskQueue<Task<unknown, unknown>>();
+    this.taskQueue = new TaskQueue<Task>();
     this.agreementPoolService = new AgreementPoolService(yagnaApi, this.options);
     this.paymentService = new PaymentService(yagnaApi, this.options);
     this.marketService = new MarketService(this.agreementPoolService, yagnaApi, this.options);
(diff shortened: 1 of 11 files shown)


src/executor/executor.ts
```diff
@@ -117,7 +117,7 @@ export class TaskExecutor {
   private networkService?: NetworkService;
   private statsService: StatsService;
   private activityReadySetupFunctions: Worker[] = [];
-  private taskQueue: TaskQueue<Task<unknown, unknown>>;
+  private taskQueue: TaskQueue<Task>;
```
grisha87, line 120: One nice improvement can be to change the signature to `TaskQueue<T extends QueueableTask = Task>` to provide a nice default Then we could just use `private taskQueue: TaskQueue;` instead.

src/golem_network/golem_network.ts
```diff
@@ -98,7 +98,7 @@ export class GolemNetwork {
    * console.log(status);
    * ```
    */
-  public async createJob<Output = unknown>(worker: Worker<unknown, Output>) {
+  public async createJob<Output = unknown>(worker: Worker<Output>) {
```
grisha87, line 101: I think that we could drop `= unknown` as the type inference would work here just perfectly.

tests/unit/tasks_service.test.ts
```diff
@@ -2,18 +2,18 @@ import * as activityMock from "../mock/rest/activity";
...
     logger.clear();
     activityMock.clear();
-    queue = new TaskQueue<Task<any, any>>();
+    queue = new TaskQueue<Task>();
   });
   it("should process new task in queue", async () => {
-    const worker: Worker<null, Result> = async (ctx) => ctx.run("some_shell_command");
-    const task = new Task<null, Result>("1", worker);
+    const worker: Worker<Result> = async (ctx) => ctx.run("some_shell_command");
+    const task = new Task<Result>("1", worker);
```
grisha87, line 16: Suggestion: Write code to promote type inference. The equivalent would be:

```ts
const worker = async (ctx: WorkContext) => ctx.run("some_shell_command");
const task = new Task("1", worker);
```

This way We don't have to specify `Worker<Result>` | `Task<Result>` and the type inference does the job.

tests/unit/work.test.ts
```diff
@@ -17,7 +17,7 @@ describe("Work Context", () => {
   describe("Executing", () => {
     it("should execute run command", async () => {
       const activity = await Activity.create("test_agreement_id", yagnaApi);
-      const worker: Worker<void, Result> = async (ctx) => ctx.run("some_shell_command");
+      const worker: Worker<Result> = async (ctx) => ctx.run("some_shell_command");
```
grisha87, line 20: Same comment about type inference applies here. Refactored example:

```ts
const worker = async (ctx: WorkContext) => ctx.run("some_shell_command");
```

grisha87 (approved):

SewerynKras:
With the change to `executor.run`'s return type, we now get the correct type inference:
image

For reference, previously the type union always had `| undefined`
image

grisha87 (approved):

PR #678, opened 2023-11-24
feat: activities can last longer than 30 minutes

grisha87:
JST-606

This feature implements mid-agreement payments as the default behaviour of the SDK.
It has been introduced in a backward compatible manner with few symbols getting deprecated in the process.
By default, the SDK will accept debit notes each 2 minutes and issue mid-agreement payments each 12 hours.


src/market/config.ts
```diff
@@ -19,34 +19,51 @@ const DEFAULTS = {
...
 
   constructor(options?: DemandOptions) {
     this.subnetTag = options?.subnetTag || EnvUtils.getYagnaSubnet() || DEFAULTS.subnetTag;
-    this.timeout = options?.marketTimeout || DEFAULTS.marketTimeout;
-    this.expiration = options?.marketOfferExpiration || DEFAULTS.marketOfferExpiration;
+
+    // Handling demand/activity expiration in a backward compatible way
+    const oldExpiration = options?.marketOfferExpiration;
+    const newExpiration = options?.expiration;
+    this.expiration = oldExpiration || newExpiration || DEFAULTS.expiration;
```
SewerynKras, line 37: If user sets `marketOfferExpiration` or `expiration` to `0` it will be changed to `DEFAULTS.expiration`. Is this expected? Why not use `??` here?

src/market/config.ts
```diff
@@ -19,34 +19,51 @@ const DEFAULTS = {
...
+    }
+
     this.offerFetchingInterval = options?.offerFetchingInterval || DEFAULTS.offerFetchingInterval;
     this.logger = options?.logger;
     this.maxOfferEvents = options?.maxOfferEvents || DEFAULTS.maxOfferEvents;
     this.eventTarget = options?.eventTarget;
+    this.debitNotesAcceptanceTimeoutSec =
+      options?.debitNotesAcceptanceTimeoutSec || DEFAULTS.debitNotesAcceptanceTimeoutSec;
+    this.midAgreementPaymentTimeoutSec =
+      options?.midAgreementPaymentTimeoutSec || DEFAULTS.midAgreementPaymentTimeoutSec;
```
SewerynKras, line 52: Same question as above, setting these settings to `0` will unexpectedly change them to the default values

src/market/demand.ts
```diff
@@ -14,19 +14,74 @@ export interface DemandDetails {
...
+   *
+   * If your activity is about to operate for 5-30 min, {@link expiration} is sufficient.
+   *
+   * If your activity is about to operate for 30min-10h, {@link debitNotesAcceptanceTimeoutSec} should be set as well.
+   *
+   * If your activity is about to operate longer than 10h, you need set both {@link debitNotesAcceptanceTimeoutSec} and {@link midAgreementPaymentTimeoutSec}.
+   *
+   *
+   */
+  expiration?: number;
```
SewerynKras, line 53: IMO it's a good idea to include the unit of time in the option name - `expirationMs`

SewerynKras (changes requested):

src/market/demand.ts
grisha87: Good hint. I will rebase the PR to `beta` and make it BC-breaking way.

src/market/config.ts
grisha87: This value is translated to a required property of the agreement which determines the time after which the agreement will be terminated. Putting `0` here would essentially lead to a situation where demand will be rejected by all providers on the network.

Thus, I think that having a sane default here make sense, as it's going to be harder for the user to misconfigure the demand.

src/market/config.ts
grisha87: Depending on the setting:

If `debitNotesAcceptanceTimeoutSec` would be set to 0, this would make the demand get rejected by providers or in some cases limit the lifetime of an activity to 30 minutes.

If `midAgreementPaymentTimeoutSec` would be set to 0, this would make the demand get rejected by providers or in some cases limit the lifetime of an activity to 10 hours.

Now, if we will provide the defaults in these cases (like in the other one as well), I think we will be trying to be wrongly "helpful" again.

I think that I'll issue a warning and accept the user setting. I'm only afraid that most of the time the users won't be using logs with the SDK and this might not be noticed.

src/market/config.ts
pgrzy-golem: What's the behavior of expiration==0?

grisha87: @pgrzy-golem , for the demand it would mean that it should expire "now", effectively rendering it useless.

src/market/config.test.ts
```diff
@@ -0,0 +1,57 @@
...
+            () =>
+              new DemandConfig({
+                debitNotesAcceptanceTimeoutSec: value,
+              }),
+          ).toThrow("The debit note acceptance timeout time has to be a positive integer");
+        },
+      );
+    });
+
+    describe("Mig-agreement payments timeout configuration", () => {
```
pgrzy-golem, line 43: Mig or Mid?

src/market/config.ts
```diff
@@ -19,34 +19,62 @@ const DEFAULTS = {
...
   constructor(options?: DemandOptions) {
-    this.subnetTag = options?.subnetTag || EnvUtils.getYagnaSubnet() || DEFAULTS.subnetTag;
-    this.timeout = options?.marketTimeout || DEFAULTS.marketTimeout;
-    this.expiration = options?.marketOfferExpiration || DEFAULTS.marketOfferExpiration;
-    this.offerFetchingInterval = options?.offerFetchingInterval || DEFAULTS.offerFetchingInterval;
     this.logger = options?.logger;
-    this.maxOfferEvents = options?.maxOfferEvents || DEFAULTS.maxOfferEvents;
     this.eventTarget = options?.eventTarget;
+
+    this.subnetTag = options?.subnetTag || EnvUtils.getYagnaSubnet() || DEFAULTS.subnetTag;
```
pgrzy-golem, line 35: Is `||` left on purpose?

src/market/factory.test.ts
```diff
@@ -0,0 +1,55 @@
...
+import { anything, capture, imock, instance, mock, when } from "@johanblumenberg/ts-mockito";
+import { Package } from "../package";
+import { Allocation } from "../payment";
+import { YagnaApi } from "../utils";
+import { RequestorApi as MarketRequestorApi } from "ya-ts-client/dist/ya-market/src/api/requestor-api";
+
+describe("Demand Factory", () => {
+  describe("mid-agreement payments support", () => {
+    describe("default behaviour", () => {
+      test("it configures mid-agreement payments by default", async () => {
```
pgrzy-golem, line 11: We are using it() everywhere, not test(). I think keeping only one of those would be better for consistency.

src/market/config.test.ts
```diff
@@ -0,0 +1,57 @@
+import { DemandConfig } from "./config";
+
+describe("Demand Config", () => {
+  describe("Positive cases", () => {
+    test("It will accept proper config values without an error", () => {
```
pgrzy-golem, line 5: We are using it() everywhere, not test(). I think keeping only one of those would be better for consistency.

pgrzy-golem (dismissed):

src/market/demand.ts
```diff
@@ -14,19 +14,70 @@ export interface DemandDetails {
...
-  marketTimeout?: number;
-  marketOfferExpiration?: number;
+
+  /**
+   * Determines the expiration time of the offer and the resulting activity in milliseconds.
+   *
+   * The value of this field is used to define how long the demand is valid for yagna to match against.
+   * In addition, it will determine how long the resulting activity will be active.
+   *
+   * For example: if `expiration` is set to 10 minutes, the demand was created and starting an activity
```
SewerynKras, line 31: `expirationSec`

SewerynKras (approved):

PR #690, opened 2023-11-28
maxTaskRetries validation

mgordel:
The bug with `maxTaskRetries: 0`  does not exist, it was fixed somewhere earlier... I only added unit tests and e2e tests that verify this. Additionally, I added validation of the maxTaskRetries parameter. Now throws an error for a value less than zero.


grisha87:
@mgordel Please take care of making use of `BREAKING CHANGE` annotation. This change is such one, and we should include this info as the last line of the commit message. You can fix this when merging the PR by using the "squash and merge" option.

PR #693, opened 2023-11-29
docs: redesigned readme and separated info about testing into separat…

grisha87:
…e file

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ distributed, computational loads through Golem Network.
 
 ## System requirements
 
-To use `golem-js` you need to have `yagna` installed at least in version `v0.13.2`. Yagna is a service that communicates and performs operations on the Golem Network, upon your requests via the SDK. You can [follow these instructions](https://docs.golem.network/docs/creators/javascript/quickstarts/quickstart#install-yagna-2) to set it up.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -83,122 +73,39 @@ yarn build
...
 
-(async function main() {
-  const executor = await TaskExecutor.create("9a3b5d67b0b27746283cb5f287c13eab1beaa12d92a9f536b747c7ae");
+(async () => {
+  const executor = await TaskExecutor.create("golem/alpine:latest");
   await executor.run(async (ctx) => console.log(await ctx.run("echo 'Hello World'")).stdout);
   await executor.shutdown();
-})();
+})().catch((err) => console.error(err, "Error in main"));
 ```
```
mgordel, line 88: I'm just wondering if this snippet shouldn't look like `hello.mjs` from our `examples` directory?
```js
import { TaskExecutor } from "@golem-sdk/golem-js";

(async function main() {
  const executor = await TaskExecutor.create("golem/alpine:latest");
  try {
    await executor.run(async (ctx) => console.log((await ctx.run("echo 'Hello World'")).stdout));
  } catch (error) {
    console.error("Computation failed:", error);
  } finally {
    await executor.shutdown();
  }
})();
```
or maybe it's better to use top level await instead of IIFE .. (?)

cryptobench (approved):

mgordel (approved):

PR #723, opened 2023-12-14
JST-652: Implement coding conventions

grisha87:
This PR implements the agreed coding conventions:

- private members cannot be prefixed with `_`
- we don't use poperty getters/setters - switched to `setX, getX` accessors
- the primary naming style is `camelCase` with `UPPER_CASE` and `PascalCase` in other cases

### DIFF
--- a/src/golem_network/golem_network.ts
+++ b/src/golem_network/golem_network.ts
@@ -1,42 +1,39 @@
 import { v4 } from "uuid";
 import { YagnaOptions } from "../executor";
 import { Job } from "../job";
-import { Yagna } from "../utils";
+import { Yagna, YagnaApi } from "../utils";
 import { RunJobOptions } from "../job/job";
 import { GolemError } from "../error/golem-error";
+
+type GolemNetworkConfig = Partial<RunJobOptions> & { yagna?: YagnaOptions };
 
 /**
  * The Golem Network class provides a high-level API for running jobs on the Golem Network.
  */
 export class GolemNetwork {
-  private _yagna: Yagna | null = null;
+  private yagna: Yagna;
+  private api: YagnaApi | null = null;
+
+  private initializedState = false;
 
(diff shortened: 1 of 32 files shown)


mgordel (approved):
I didn't notice anything incorrect

src/golem_network/golem_network.ts
```diff
@@ -1,42 +1,39 @@
...
-      return;
-    }
-    const yagna = new Yagna(this.config.yagna);
-    // this will throw an error if yagna is not running
-    await yagna.connect();
-    this._yagna = yagna;
+    await this.yagna.connect();
+    this.api = this.yagna.getApi();
+
+    this.initializedState = true;
```
SewerynKras, line 36: Not sure we need  all three `api`, `yagna` and `initializedState`, can't we achieve the same business logic with just `api`?

```ts
public async init() {
    if (this.isInitialized()) {
      return;
    }
    const yagna = new Yagna(this.config.yagna);
    // this will throw an error if yagna is not running
    await yagna.connect();
    this.api = this.yagna.getApi();
}

public isInitialized() {
  return this.api !== null
}
```

src/golem_network/golem_network.ts
```diff
@@ -1,42 +1,39 @@
 import { v4 } from "uuid";
 import { YagnaOptions } from "../executor";
 import { Job } from "../job";
-import { Yagna } from "../utils";
+import { Yagna, YagnaApi } from "../utils";
 import { RunJobOptions } from "../job/job";
 import { GolemError } from "../error/golem-error";
 
+type GolemNetworkConfig = Partial<RunJobOptions> & { yagna?: YagnaOptions };
```
SewerynKras, line 8: good idea with the type, can we export it in the root `index.ts`?

SewerynKras (approved):

PR #735, opened 2023-12-19
fix(payments): preventing double payments resulting from a race condi…

grisha87:
…tion

Fixed the situation where invoice and debit notes sent at the end of the agreement resulted with double payments. While the original issue has to be solved in yagna, we provide an integrity ensuring solution right now.

Doubled payments can still occur, if the debit note will arrive before the invoice. Solving this was not part of this particular change.

### DIFF
--- a/src/payment/agreement_payment_process.test.ts
+++ b/src/payment/agreement_payment_process.test.ts
@@ -1,0 +1,243 @@
+import { AgreementPaymentProcess } from "./agreement_payment_process";
+import { anything, instance, mock, objectContaining, reset, verify, when } from "@johanblumenberg/ts-mockito";
+import { Agreement } from "../agreement";
+import { Allocation } from "./allocation";
+import { Invoice } from "./invoice";
+import { InvoiceStatus } from "ya-ts-client/dist/ya-payment";
+import { RejectionReason } from "./rejection";
+import { DebitNote } from "./debit_note";
+
+const agreementMock = mock(Agreement);
+const allocationMock = mock(Allocation);
+const invoiceMock = mock(Invoice);
+const debitNoteMock = mock(DebitNote);
+
+beforeEach(() => {
+  reset(agreementMock);
+  reset(allocationMock);
+  reset(invoiceMock);
+  reset(debitNoteMock);
+});
+
+describe("AgreementPaymentProcess", () => {
+  describe("Accepting Invoices", () => {
+    describe("Positive cases", () => {
+      it("accepts a single invoice", async () => {
+        when(allocationMock.id).thenReturn("1000");
+        when(invoiceMock.amount).thenReturn("0.123");
+
+        const process = new AgreementPaymentProcess(instance(agreementMock), instance(allocationMock), {
+          debitNoteFilter: () => true,
+          invoiceFilter: () => true,
+        });
+
+        const success = await process.addInvoice(instance(invoiceMock));
+
+        expect(success).toEqual(true);
+        verify(invoiceMock.accept("0.123", "1000")).called();
+        expect(process.isFinished()).toEqual(true);
+      });
+
+      it("rejects invoice if it's ignored by the user defined invoice filter", async () => {
+        when(allocationMock.id).thenReturn("1000");
+        when(invoiceMock.amount).thenReturn("0.123");
+
+        const process = new AgreementPaymentProcess(instance(agreementMock), instance(allocationMock), {
+          debitNoteFilter: () => true,
+          invoiceFilter: () => false,
+        });
+
+        const success = await process.addInvoice(instance(invoiceMock));
+
+        expect(success).toEqual(false);
+        verify(
+          invoiceMock.reject(
+            objectContaining({
+              rejectionReason: RejectionReason.RejectedByRequestorFilter,
+              message: "Invoice rejected by Invoice Filter",
+              totalAmountAccepted: "0",
+            }),
+          ),
+        ).called();
+        expect(process.isFinished()).toEqual(false);
+      });
+
+      it("accepts the duplicated invoice if the previous one is still not processed", async () => {
+        when(allocationMock.id).thenReturn("1000");
+        when(invoiceMock.amount).thenReturn("0.123");
+
+        const process = new AgreementPaymentProcess(instance(agreementMock), instance(allocationMock), {
+          debitNoteFilter: () => true,
+          invoiceFilter: () => true,
+        });
+
+        const invoice = instance(invoiceMock);
+
+        // Simulate issue with accepting the first one
+        const issue = new Error("Failed to accept in yagna");
+        when(invoiceMock.accept("0.123", "1000"))
+          .thenReject(issue) // On first call
+          .thenResolve(); // On second call
+
+        await expect(() => process.addInvoice(invoice)).rejects.toThrow(issue);
+
+        // Then simulate the duplicate coming again
+        const success = await process.addInvoice(invoice);
+
+        expect(success).toEqual(true);
+        verify(invoiceMock.accept("0.123", "1000")).twice();
+        expect(process.isFinished()).toEqual(true);
+      });
+    });
+
+    describe("Negative cases", () => {
+      it("doesn't accept the same invoice twice if the previous one was already processed", async () => {
+        // TODO: False and no error to not break?
+        when(invoiceMock.getStatus()).thenResolve(InvoiceStatus.Accepted);
+
+        const process = new AgreementPaymentProcess(instance(agreementMock), instance(allocationMock), {
+          debitNoteFilter: () => true,
+          invoiceFilter: () => true,
+        });
+
+        const invoice = instance(invoiceMock);
+
+        const success = await process.addInvoice(invoice);
+
+        expect(success).toEqual(true);
+        expect(process.isFinished()).toEqual(true);
+        await expect(() => process.addInvoice(invoice)).rejects.toThrow(
+          "This agreement is already covered with an invoice",
+        );
+      });
+    });
+  });
+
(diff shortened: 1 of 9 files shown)


tests/unit/payment_service.test.ts
```diff
@@ -109,16 +107,16 @@ describe("Payment Service", () => {
...
         logger,
         invoiceFilter: alwaysRejectInvoiceFilter,
+        paymentTimeout: TEST_PAYMENT_TIMEOUT_MS,
       });
       setExpectedEvents(invoiceEvents);
       setExpectedInvoices(invoices);
       await paymentService.createAllocation();
-      await paymentService.run();
       paymentService.acceptPayments(agreement);
-      await new Promise((res) => setTimeout(res, 200));
```
grisha87, line 118: We could remove the other occurrences as well.

tests/unit/payment_service.test.ts
```diff
@@ -109,16 +107,16 @@ describe("Payment Service", () => {
...
+      await paymentService.run();
       await logger.expectToInclude(
         `Invoice has been rejected for provider ${agreement.provider.name}. Reason: Invoice rejected by Invoice Filter`,
-        100,
+        1_000,
```
grisha87, line 119: Could be updated in other places as well.

src/payment/service.ts
```diff
@@ -103,91 +102,105 @@ export class PaymentService {
...
+  // Reason: We will remove this in 2.0
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  acceptDebitNotes(_agreementId: string) {
+    return;
   }
 
-  acceptDebitNotes(agreementId: string) {
-    this.agreementsDebitNotes.add(agreementId);
+  private getNumberOfUnpaidAgreements() {
+    const unpaidProcesses = [...this.processes.values()].filter((p) => !p.isFinished());
```
grisha87, line 141: unfinishedProcesses

src/payment/service.ts
```diff
@@ -103,91 +102,105 @@ export class PaymentService {
...
+      ),
+    );
+  }
+
+  /**
+   * @deprecated, Use `acceptPayments` instead
+   */
+  // Reason: We will remove this in 2.0
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  acceptDebitNotes(_agreementId: string) {
```
grisha87, line 136: Log a warning here

src/payment/rejection.ts
```diff
@@ -11,6 +11,10 @@ export enum RejectionReason {
    * as the payment will be already made when we accept the invoice.
    */
   NonPayableAgreement = "NON_PAYABLE_AGREEMENT",
+  Duplicate = "DUPLICATE",
```
grisha87, line 14: Tidy up the dictionary from the things that are not used.

src/payment/agreement_payment_process.ts
```diff
@@ -0,0 +1,158 @@
...
+    };
+    await debitNote.reject(reason);
+    this.logger?.warn(`DebitNote has been rejected for agreement ${debitNote.agreementId}. Reason: ${reason.message}`);
+  }
+
+  private async applyInvoice(invoice: Invoice) {
+    const isAlreadyFinalized = await this.hasProcessedInvoice();
+
+    if (isAlreadyFinalized) {
+      throw new Error("This agreement is already covered with an invoice");
```
grisha87, line 118: @mgordel , this is something that I'm reconsidering. When we process debit notes, when a duplicate comes, we warn + ignore.

Here, when a duplicate of the invoice comes, we throw - I don't like that. Perhaps: if that's a duplicate (id===id) then ignore, but when that's a new invoice, then throw (as that's an error situation). 

What's your opinion?

mgordel: maybe we should separate the case when a new one came:
 1. the previous one was accepted or settled -> we should reject the new one and/or throw error (?)
 2. the previous one was rejected -> we should try to process the new one (?)

grisha87: The invoice is firstly recorded by yagna in "received" state and then passed to us when we poll for it. In context of that:

> the previous one was accepted or settled -> we should reject the new one and/or throw error (?)

Given a duplicate arrives, this would mean, that we "accepted or settled" it previously. In this case I wouldn't reject it, I would neither alarm the user (no error report), nor throw an error - handle this idempotently, issue a warning.

> the previous one was rejected -> we should try to process the new one (?)

Given a duplicate arrives, this would mean that we "rejected" it previously. If I had a reason to reject it, then I shouldn't challenge that on the second call. Idempotently warn about the situation, not an error log or Error throw.

In general: if I have a duplicate and there was a decision about it (status !== RECEIVED), then I should not try to alter that decision. On the other hand if that's a duplicate event and `status === RECEIVED`, then that's a chance to make the decision - we should process this invoice/debit-note.

---

So the only case that remains is that I'm getting another invoice for any reason (like a bug) after I already got one (no duplicate compared to the old one, but a different one).

That's breaking the rule of a single invoice per agreement. If it breaks the rule, I shouldn't try to process the new invoice regardless of the state of the previous one (accepted/settled/received/rejected). In such case, this calls for a rejection of the new invoice as this might be an attempt to force payments on the requestor. (I'm a malicious provider, I have an agreement, and then I'm trying to exploit that by sending 10 invoices for the same agreement). Rule: the first invoice that gets to us, is the one and only. The rest will be rejected.

src/payment/agreement_payment_process.ts
```diff
@@ -0,0 +1,158 @@
...
+        return false;
+      }
+    }
+
+    this.debitNotes.set(debitNote.id, debitNote);
+
+    const acceptedByFilter = await this.filters.debitNoteFilter(debitNote.dto);
+
+    if (acceptedByFilter) {
+      await debitNote.accept(debitNote.totalAmountDue, this.allocation!.id);
```
grisha87, line 84: !

mgordel: allocation is always not null in this context

src/payment/agreement_payment_process.test.ts
```diff
@@ -0,0 +1,242 @@
...
+
+        expect(success).toEqual(true);
+        verify(invoiceMock.accept("0.123", "1000")).twice();
+        expect(process.isFinished()).toEqual(true);
+      });
+    });
+
+    describe("Negative cases", () => {
+      it("doesn't accept the same invoice twice if the previous one was already processed", async () => {
+        // TODO: False and no error to not break?
```
mgordel, line 95: we should implement it as we agreed above

src/payment/agreement_payment_process.ts
mgordel: ok, I agree with this approach, assuming that the duplicate is the same invoice with the same ID, we proceed as you described.

However, receiving more than one invoice per agreemnet is considered a fraud attempt and we treat it as a scam.

It makes sense.

grisha87: I've implemented all of the established rules:

```
 PASS  src/payment/agreement_payment_process.test.ts
  AgreementPaymentProcess
    Accepting Invoices
      Basic use cases
        ✓ accepts a invoice in RECEIVED state (4 ms)
        ✓ rejects invoice if it's ignored by the user defined invoice filter (1 ms)
        ✓ throws an error when invoice in a state different than RECEIVED (7 ms)
      Dealing with duplicates
        ✓ accepts the duplicated invoice if accepting the previous one failed (9 ms)
        ✓ ignores the duplicate if the original invoice has been already decided upon
        ✓ accepts the duplicate if the original invoice has not been already decided upon (still in RECEIVED state) (8 ms)
        ✓ doesn't accept the same invoice twice if the previous one was already processed (1 ms)
      Security
        ✓ throws an error when there's already an invoice present, and a different one is passed to the process (1 ms)
    Accepting DebitNotes
      Basic use cases
        ✓ accepts a single debit note (1 ms)
        ✓ rejects debit note if it's ignored by the user defined debit note filter
        ✓ rejects debit note if there is already an invoice for that process
      Dealing with duplicates
        ✓ accepts the duplicated debit note if accepting the previous failed (1 ms)
        ✓ doesn't accept the same debit note twice if the previous one was already processed (1 ms)
```

src/payment/agreement_payment_process.ts
```diff
@@ -0,0 +1,166 @@
...
+        debitNote,
+        RejectionReason.RejectedByRequestorFilter,
+        "DebitNote rejected by DebitNote Filter",
+      );
+
+      return false;
+    }
+
+    await debitNote.accept(debitNote.totalAmountDue, this.allocation.id);
+    this.logger?.debug(`DebitNote accepted for agreement ${debitNote.agreementId}`);
```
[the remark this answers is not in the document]
grisha87: I think we could even do it in all errors.

src/payment/agreement_payment_process.test.ts
```diff
@@ -0,0 +1,344 @@
...
+        const firstSuccess = await process.addInvoice(invoice);
+        const secondSuccess = await process.addInvoice(invoice);
+
+        expect(firstSuccess).toEqual(true);
+        expect(secondSuccess).toEqual(true);
+        verify(invoiceMock.accept("0.123", "1000")).twice();
+        expect(process.isFinished()).toEqual(true);
+      });
+
+      it("doesn't accept the same invoice twice if the previous one was already processed", async () => {
```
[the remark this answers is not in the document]
grisha87: No, it's a duplicate and even my IDE notices the duplicated code :)

mgordel (approved):
Now it's great. Good job. And we have a complete example of how to refactor our unit tests using ts-mockito

PR #755, opened 2024-01-09
JST-336/fix logging

SewerynKras:
BREAKING CHANGE: The SDK won't log anything to the console by default. Please refer to the `debug` package's README to learn more.


SewerynKras:
https://github.com/golemfactory/golem-js/actions/runs/7475569013
✔️ e2e tests
✔️ cypress
✔️ examples

grisha87 (changes requested):

tests/cypress/support/e2e.ts
```diff
@@ -15,3 +15,10 @@
 //
 // // Import commands.js using ES2015 syntax:
 // import "./commands";
+
+beforeEach(() => {
+  cy.intercept("GET", "https://unpkg.com/@golem-sdk/golem-js", (req) => {
```
grisha87, line 20: Nice:D

src/task/work.ts
```diff
@@ -195,40 +202,40 @@ export class WorkContext {
    * @param options Additional run options.
    */
   async transfer(from: string, to: string, options?: CommandOptions): Promise<Result> {
-    this.logger.debug(`WorkContext: transfering ${from} to ${to}`);
+    this.logger.debug(`Transferring ${from} to ${to}`);
```
grisha87, line 205: We might push `from` and `to` into context so that the template will not be always "stringified" by V8. Same remark to the other places where the teamplate is used. That's probably a minor performance improvement.

src/storage/ws-browser.ts
```diff
@@ -138,7 +138,7 @@ export class WebSocketBrowserStorageProvider implements StorageProvider {
       const serviceId = this.services.get(url);
       if (serviceId) {
         this.deleteService(serviceId).catch((e) =>
-          this.logger.warn(`[WebSocketBrowserStorageProvider] Failed to delete service ${serviceId}: ${e}`),
+          this.logger.error(`[WebSocketBrowserStorageProvider] Failed to delete service ${serviceId}: ${e}`),
```
grisha87, line 141: Please restore the logging level and push `serviceId` and `error` into the context.

src/payment/payments.ts
```diff
@@ -125,13 +124,14 @@ export class Payments extends EventTarget {
...
-          );
+          this.logger.debug("New Debit Note received", {
+            agreementId: debitNote.agreementId,
+            amount: debitNote.totalAmountDue,
+          });
         }
       } catch (error) {
         const reason = error.response?.data?.message || error;
-        this.logger?.debug(`Unable to get debit notes. ${reason}`);
+        this.logger.error(`Unable to get debit notes.`, { reason });
```
grisha87, line 134: I'm not sure if that shouldn't be a warning. I'll see the UX when testing it

src/payment/payments.ts
```diff
@@ -80,11 +76,15 @@ export class Payments extends EventTarget {
...
+            id: invoice.id,
+            agreementId: invoice.agreementId,
+            amount: invoice.amount,
+          });
         }
       } catch (error) {
-        const reason = error.response?.data?.message || error;
-        this.logger?.debug(`Unable to get invoices. ${reason}`);
+        const reason = error.response?.data?.message || error.message || error;
+        this.logger.error(`Unable to get invoices.`, { reason });
```
grisha87, line 87: I'm not sure if that shouldn't be a warning. I'll see the UX when testing it

src/network/service.ts
```diff
@@ -38,6 +38,6 @@ export class NetworkService {
 
   async end() {
     await this.network?.remove();
-    await this.logger?.debug("Network Service has been stopped");
+    this.logger.debug("Network Service has been stopped");
```
grisha87, line 41: To consider: Places where the core services are starting/stopping, could be reported on the info level so that we as developers see these operations without diving into `debug`. I't like with starting regular apps, things like "starting service" "stopping service" are usually on the INFO level.

tests/unit/network_service.test.ts
```diff
@@ -11,7 +11,15 @@ describe("Network Service", () => {
     it("should start service and create network", async () => {
       const networkService = new NetworkService(yagnaApi, { logger });
       await networkService.run("test_owner_id");
-      await logger.expectToMatch(/Network created: ID: .*, IP: 192.168.0.0, Mask: 255.255.255.0/, 10);
+      await logger.expectToInclude(
+        "Network created",
+        {
```
grisha87, line 16: Damn, I love this change!

grisha87 (changes requested):

src/payment/payments.ts
grisha87: OK, so I've checked and the result is as follows:

```
[09:52:00.585] ERROR (12208): Unable to get invoices.
    reason: "socket hang up"
```

But we can actually leave it on the error level and fix the original issue which is the broken logic of gracefully shutting down the services in the SDK, that lead to the abrupt socket hang up.

src/payment/payments.ts
grisha87: This can stay, as per: https://github.com/golemfactory/golem-js/pull/755#discussion_r1452059163

src/network/service.ts
SewerynKras: Good idea

grisha87 (approved):

PR #756, opened 2024-01-10
fix: changed the default testnet to 'holesky' from 'goerli'

grisha87:


src/market/proposal.ts
```diff
@@ -13,7 +13,7 @@ export type ProposalProperties = Record<string, string | number | string[] | num
   "golem.activity.caps.transfer.protocol": string[];
   "golem.com.payment.debit-notes.accept-timeout?": number;
   "golem.com.payment.platform.erc20-polygon-glm.address"?: string;
-  "golem.com.payment.platform.erc20-goerli-tglm.address"?: string;
+  "golem.com.payment.platform.erc20-holesky-tglm.address"?: string;
```
mgordel, line 16: Has goerli been replaced by holesky? or maybe holesky is another tetsnet that can be used? if both testnets are available, there should be an additional property instead of replacement..

grisha87: https://www.alchemy.com/blog/goerli-faucet-deprecation

Goerli is going to be shut down this January, I'd say that's a proper replacement :)

mgordel (approved):
one last thing, does the `yagna payment fund` also fund the holesky account by default? because if not, we probably need to fix it in the script for funding the requestor in e2e tests..

PR #762, opened 2024-01-15
Outbound example fixes

mgordel:
The previous example did not run properly and required a provider with a whitelisted url to coingecko.com to work properly. I changed the example to use the url from ipfs.io, which is whitelisted by default.

Additionally, I fixed bugs with the ProviderInfo attribute in the ProposalRejected event and the stats service.

### DIFF
--- a/examples/external-request/request.ts
+++ b/examples/external-request/request.ts
@@ -1,32 +1,36 @@
 import { TaskExecutor } from "@golem-sdk/golem-js";
-import { readFileSync } from "fs";
+import { readFile } from "fs/promises";
 import { fileURLToPath } from "url";
 const DIR_NAME = fileURLToPath(new URL(".", import.meta.url));
 
 (async function main() {
+  // Load the manifest.
+  const manifest = await readFile(`${DIR_NAME}/manifest.json`);
+
+  // Create and configure a TaskExecutor instance.
   const executor = await TaskExecutor.create({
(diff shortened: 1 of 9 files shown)


examples/external-request/request.ts
```diff
@@ -1,32 +1,36 @@
...
-    console.error("Computation failed:", error);
+    const url = "https://ipfs.io/ipfs/bafybeihkoviema7g3gxyt6la7vd5ho32ictqbilu3wnlo3rs7ewhnp7lly";
+    const results = await executor.run(async (ctx) =>
+      ctx
+        .beginBatch()
+        .run(`curl ${url} -o /golem/work/example.jpg`)
+        .downloadFile("/golem/work/example.jpg", `${DIR_NAME}/example.jpg`)
+        .end(),
+    );
+    console.log(results);
```
SewerynKras, line 31: I would rather log something more readable to the user than the raw results object, how about:
```ts
    if (results[1].result === "Ok") {
      console.log("Downloaded file to", `${DIR_NAME}/example.jpg`);
    } else {
      console.error("Something went wrong", results[0].result);
    }
```
What do you think @mgordel ?

mgordel: yes, you're right, this is a more readable form

SewerynKras (approved):
lgtm

PR #764, opened 2024-01-16
JST-664: Add a way to manually list and pay invoices

SewerynKras:


src/payment/InvoiceProcessor.ts
```diff
@@ -0,0 +1,286 @@
...
+      success: true;
+      amount: string;
+      dryRun: boolean;
+    }
+  | {
+      invoiceId: string;
+      allocation: Allocation;
+      success: false;
+      amount: string;
+      reason: unknown;
```
grisha87, line 19: Both variants of the type differ in the `reason` field this could be unified to a single type with `reason?:unknown`

grisha87 (approved):

src/payment/InvoiceProcessor.ts
SewerynKras: `reason` is only defined if `success: false` so IMO a discriminated union like this one works best here

PR #774, opened 2024-01-24
fix(network): fixed adding an existing node to the network

mgordel:
Fixed a bug related to adding a node to the network. There was no checking whether the node existed. Additionally, added the ability to remove a node from the network and handle it in TaskService.

### DIFF
--- a/src/network/network.ts
+++ b/src/network/network.ts
@@ -49,6 +49,7 @@ export class Network {
   private gateway?: IPv4;
   private nodes = new Map<string, NetworkNode>();
   private logger: Logger;
+  private releasedIpAddresses: Array<AbstractIPNum> = [];
 
   /**
    * Create a new VPN.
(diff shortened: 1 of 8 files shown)


src/network/network.ts
```diff
@@ -140,6 +141,33 @@ export class Network {
...
   }
 
+  /**
+   * Remove the node from the network
+   * @param nodeId
+   */
+  async removeNode(nodeId: string): Promise<void> {
+    const node = this.nodes.get(nodeId);
+    if (!node) {
+      throw new GolemError(`Unable to remove node ${nodeId}. There is no such node in the network`);
```
grisha87, line 151: I would adjust the type after you merge the PR with new error structures.

tests/unit/network.test.ts
```diff
@@ -99,6 +99,28 @@ describe("Network", () => {
       await expect(network.addNode("4")).rejects.toThrow("No more addresses available in 192.168.0.0/30");
     });
 
+    it("should add a node if the network was full and some node was removed", async () => {
```
grisha87, line 102: The subject looks missleading. When I reed the test, I consider "should re-use an IP address when a node previously using it leaves the network".

I'm not sure if I like the idea of re-using the IP for different nodes. Lets consider that the ID of the node is a provider or a requestor on such a network. Lets assume that providers and requetsors are "blinking" on the network - going in and out.

The typicall assumption with IPs is that you have the IP "leased" in a network for some period of time so that you retain it even if you blink shortly. Here we're trying to re-use that, which I don't think is a good idea.

- I create a network
- I put providers A, B, C in the network
- Tey get IPs: A.1, B.2, C.3
- The provider A.1 gets removed from the network
- D joins network as D.1

Won't it be the case that the providers B and C will still think that .1 means A and nod D?

grisha87 (approved):

grisha87 (changes requested):
As discussed during our call. Instead of saving the world when we reach the end of the address space, we should rather throw an error towards the user.

src/network/network.ts
mgordel: Yes, after the merge I will resolve all conflicts and improve it.

tests/unit/network.test.ts
```diff
@@ -99,14 +99,12 @@ describe("Network", () => {
       await expect(network.addNode("4")).rejects.toThrow("No more addresses available in 192.168.0.0/30");
     });
 
-    it("should add a node if the network was full and some node was removed", async () => {
+    it("should not add a node to a previously assigned address", async () => {
```
grisha87, line 102: Please use positive logic and tell the story in the description: "should throw an error when there are no free IPs available"

grisha87 (approved):

grisha87 (approved):

PR #775, opened 2024-01-25
fix(task-executor): propagate issues with invoice and debit note filt…

grisha87:
…er to user

Right now the payment service has silently tolerated user triggered errors from the invoice or debit note filters - the buggy code was still operating producing bad market outcomes - leading to the providers terminating agreements due to not accepted debit notes. The users were left clueless.

What the PR changes is: given there's any issue with processing an invoice or debit note, then it's not only logged in the SDK on error level, but also an `error` event is raised from the `PaymentService`. This error is then listened to by the `TaskExecutor`, which cancels itself in response - producing the information to the user and stopping the buggy code to operate.

One side effect here is that
_any_ issue with the processing will lead to the shutdown of the TaskExecutor.

BREAKING CHANGE: The TaskExecutor will now be cancelled when an issue will be raised with processing the payments - including errors in the invoice and debit note filters.

### DIFF
--- a/src/executor/executor.ts
+++ b/src/executor/executor.ts
@@ -258,7 +258,12 @@ export class TaskExecutor {
       this.statsService.run(),
       this.storageProvider?.init(),
     ]).catch((e) => this.handleCriticalError(e));
+
+    // Start listening to issues reported by the services
+    this.paymentService.events.on("error", (e) => this.handleCriticalError(e));
+
     this.taskService.run().catch((e) => this.handleCriticalError(e));
+
     if (runtimeContextChecker.isNode) this.installSignalHandlers();
     this.options.eventTarget.dispatchEvent(new Events.ComputationStarted());
     this.logger.info(`Task Executor has started`, {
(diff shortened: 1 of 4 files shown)


src/executor/executor.ts
```diff
@@ -454,19 +479,12 @@ export class TaskExecutor {
...
-        tasksInProgress: this.taskQueue.size,
-      });
-      this.isCanceled = true;
-      await this.shutdown();
-    } catch (error) {
-      this.logger.error(`Error while cancelling the executor`, error);
-    }
+  private handleCriticalError(e: Error) {
+    this.options.eventTarget?.dispatchEvent(new Events.ComputationFailed({ reason: e.toString() }));
+    this.logger.error("Critical error", e);
```
grisha87, line 484: Consider changing this to a standard console.error to hint the user about the issue even if they didn’t configure logging on TE

mgordel (approved):
Ok for me. Just one question. In addition to payment filters, the user can inject their own functions for the agreemnetSelector and proposalFilter, and in these cases maybe we should also handle it in a similar way. Now only an error will be logged if the user provides his own function with an error (for proposalFilter and agreemnetSelector)?

grisha87:
> Ok for me. Just one question. In addition to payment filters, the user can inject their own functions for the agreemnetSelector and proposalFilter, and in these cases maybe we should also handle it in a similar way. Now only an error will be logged if the user provides his own function with an error (for proposalFilter and agreemnetSelector). I'm just wondering, what do you think about this?

I think it's broadening the scope of this PR, but it's a good idea to address these spots as well. I'll introduce necessary changes and I'll update the PR.

PR #776, opened 2024-01-26
Remove deprecated method

grisha87:
BREAKING CHANGE: `PaymentService.acceptDebitNote` is no longer available, use `acceptPayments` instead.

### DIFF
--- a/examples/docs-examples/examples/transferring-data/transfer-data-in-browser.html
+++ b/examples/docs-examples/examples/transferring-data/transfer-data-in-browser.html
@@ -124,7 +124,7 @@
           })
           .catch((e) => logger.error(e));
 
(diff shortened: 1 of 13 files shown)


mgordel:
We have a few more places tagged deprecated. We could also review all of them in this PR and remove them with the BC commit.. ?

E.g. in `task/work.ts` -> `rejectResult`

grisha87:
I thin

> We have a few more places tagged deprecated. We could also review all of them in this PR and remove them with the BC commit.. ?
> 
> E.g. in `task/work.ts` -> `rejectResult`

I thik that's a good idea. Let me browse through the code :)

mgordel (approved):

PR #786, opened 2024-01-31
fix(task): fixed retrying and timeout measurement for task

mgordel:
Now the task timeout is measured for each attempt separately.
Additionally, I refactored the task service by separating individual blocks into startTask, retryTask and stopTask. They are triggered by a change in the task state.
I added another state for the task - `queued`. Now, when we start the task, the task changes its state from new -> queued and when agreement and activity are created, the task starts (state pending) and from that moment the timeout is counted. It seems that this approach is better, because now the duration of the task is counted from the moment of running the first activity command (deploy) and this does not include negotiations with the market.

### DIFF
--- a/src/activity/activity.ts
+++ b/src/activity/activity.ts
@@ -247,10 +247,14 @@ export class Activity {
               },
             );
             retryCount = 0;
+            if (!isRunning()) {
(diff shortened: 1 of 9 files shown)


SewerynKras (approved):
LGTM, tested it locally and it seems to fix the issue with task timeout not resetting between attempts

PR #788, opened 2024-02-01
fix(agreement_pool): added agreement pool size limit

mgordel:
Added a mechanism for checking whether the pool is full when releasing an agreement.
By default, the agreementMaxPoolSize parameter is set by TaskExecutor with a value equal to maxParallelTasks

### DIFF
--- a/src/agreement/config.ts
+++ b/src/agreement/config.ts
@@ -9,6 +9,7 @@ const DEFAULTS = {
   agreementSelector: randomAgreementSelectorWithPriorityForExistingOnes(),
   agreementMaxEvents: 100,
   agreementEventsFetchingIntervalSec: 5,
+  agreementMaxPoolSize: 5,
 };
 
 /**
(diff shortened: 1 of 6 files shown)


src/agreement/config.ts
```diff
@@ -35,12 +36,14 @@ export class AgreementConfig {
...
   readonly agreementSelector: AgreementSelector;
   readonly agreementMaxEvents: number;
+  readonly agreementMaxPoolSize: number;
   readonly agreementEventsFetchingIntervalSec: number;
 
   constructor(options?: AgreementServiceOptions) {
     super(options);
     this.agreementSelector = options?.agreementSelector ?? DEFAULTS.agreementSelector;
     this.agreementMaxEvents = options?.agreementMaxEvents ?? DEFAULTS.agreementMaxEvents;
+    this.agreementMaxPoolSize = options?.agreementMaxPoolSize ?? DEFAULTS.agreementMaxPoolSize;
```
SewerynKras, line 46: What is the benefit of setting `agreementMaxPoolSize` to anything other than the same value as `maxParallelTasks`? i think these two config values should just be merged to one

mgordel: In my opinion, AgreementPool does not know what a task is and does not know the concept of maxParallelTasks. In the context of agreemnt domain, it is limited only to the pool size parameter. The TaskExecutor maps maxParallelTask to the max size of the agreement pool. And since you're asking, a comment explaining why would probably be useful here. so I'll add a comment here.

We will soon get rid of TaskExecutor from this repo and whatever uses AgreementPool should set a parameter related to its context.

Maybe agreementMaxPoolSize is not the best name, but e.g. if it will be used by a service or deployment (in our new API), maxParallelTask will not fit..

SewerynKras: You're right, when reading the PR I missed that the default is `DEFAULTS.maxParallelTasks`. You can disregard this comment then

SewerynKras (approved):

PR #789, opened 2024-02-01
JST-683: Golem inspect

SewerynKras:


src/utils/yagna/activity.ts
```diff
@@ -0,0 +1,17 @@
+import { RequestorStateApi } from "ya-ts-client/dist/ya-activity";
+import { GolemPlatformError } from "../../error/golem-error";
+import { Agreement } from "ya-ts-client/dist/ya-market";
+
+export class RequestorApi extends RequestorStateApi {
+  async getActivityAgreementId(activityId: string): Promise<Agreement["agreementId"]> {
+    const res = await fetch(this.basePath + "/activity/" + activityId + "/agreement", {
```
mgordel, line 7: For all original API calls (coming from ya-ts-clinet) we have a global errorHandler set (as an interceptor to axios). If you use fetch here, you should add a catch to it that will throw GolemPlatformError by setting the property previous, e.g.
```ts
const res = await fetch( .... ).catch((e) => {
      throw new GolemPlatformError(`Failed to get activity agreement: ${e}`, e);
});
```
or try/catch if you prefer

mgordel (changes requested):

PR #804, opened 2024-02-15
Task/jst 734/update docs for better logging

grisha87:


mgordel:
Are we sure that we don't want to have any example with pure JS (.mjs or .cjs) ?

mgordel (approved):

PR #806, opened 2024-02-16
Golem Proxy

mgordel:


src/task/process.ts
```diff
@@ -2,10 +2,16 @@ import { Readable, Transform } from "stream";
 import { Activity, Result } from "../activity";
 import { GolemWorkError, WorkErrorCode } from "./error";
 import { GolemTimeoutError } from "../error/golem-error";
+import { EventEmitter } from "eventemitter3";
 
 const DEFAULTS = {
   exitWaitingTimeout: 20_000,
 };
+
+interface RemoteProcessEvents {
```
grisha87, line 11: Cleanup

src/utils/logger/logger.ts
```diff
@@ -1,11 +1,11 @@
 export interface Logger {
   child(namespace: string): Logger;
   debug(msg: string): void;
-  debug(msg: string, ctx: Record<string, unknown> | Error): void;
+  debug(msg: string, ctx: Record<string, unknown> | Error | unknown): void;
```
grisha87, line 4: Check if still required change

examples/proxy/proxy.ts
```diff
@@ -0,0 +1,72 @@
...
+        if (isClosing) {
+          console.log("Already closing, ignoring subsequent shutdown request");
+          return;
+        }
+
+        isClosing = true;
+
+        console.log("Shutting down gracefully");
+        await proxy.close();
+        await server.waitForExit();
```
mgordel, line 59: by default, after 30 seconds it will throw a `TimeoutError` because the `server` (process) will not end. We cannot send any `SIGTERM` or `SIGINT` signal to the process (on the provider/batch in activity), so the only way is to try to end the `task/workerFunction` without waiting for exit from remoteProcess. Then `TaskExecutor` will treat the task as completed and will terminate the activity and agreement. It's not a pretty solution, but I think it's the only possible one at this moment..

examples/proxy/server.js
```diff
@@ -0,0 +1,39 @@
...
+      if (err) {
+        console.error("Server close encountered an issue", err);
+      } else {
+        console.log("Server closed successfully");
+      }
+    });
+  };
+
+  process.on("SIGINT", shutdown);
+  process.on("SIGTERM", shutdown);
```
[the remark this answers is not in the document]
grisha87: I'm going to check this with the core team as lack of it sounds a bit worrying.

examples/proxy/proxy.ts
grisha87: I see your point. I've changed this line to leave a comment explaining that it's safe to leave the function after `await proxy.close()` to not run into this issue.

I've checked ya-client spec and it appears that we really can't interrupt a running batch via the API. On one hand this could be a feature request to the core team, but on the other, I think that in this context `WorkContext.spawn` is very poorly named and super misleading as it does not have a lot in common with the reality which it intended to capture. We can't even run a second command concurrently after we use `spawn`.

Thus, I think the last change which we could introduce here is a rename of this method and deprecation of the old name so that it can be removed in v3 or later.

```
WorkContext.run
WorkContext.runAndStream -> replaces: WorkContext.spawn
```

This way the `on('close', () => {})` handlers etc can still be used intuitively, and we don't give any false hopes to `spawn` users.

grisha87 (approved):

PR #808, opened 2024-02-17
docs: updated web examples to show how to debug in browser

grisha87:


.github/workflows/release.yml
```diff
@@ -115,7 +115,7 @@ jobs:
         uses: ./.github/actions/prepare-tests
 
       - name: Run web server
-        run: docker exec -t -d docker-requestor-1 /bin/sh -c "cd /golem-js/examples/web && node app.mjs"
+        run: docker exec -t -d docker-requestor-1 /bin/sh -c "cd /golem-js/examples && npm run web"
```
mgordel, line 118: using `serve` is ok, but we need to solve the problem with tests for cypress examples for docs, which were previously served by the old `app.mjs` .. 

maybe for testing we will need to copy these examples from `docs-example` to `web/dist`.. or somehow configure `serve` to serve files from `docs-example`?

mgordel:
Moved to https://github.com/golemfactory/golem-sdk-task-executor/pull/54

PR #813, opened 2024-02-21
feat: experimental reputation client

pgrzy-golem:
Experimental client for the WIP reputation system.

The client implements only a proposal filter for now and two examples. First one is a basic hello world example, the other runs a lot of tasks with no reputation system and with differently configured reputation systems.


src/experimental/reputation/system.ts
```diff
@@ -0,0 +1,233 @@
...
+ *
+ * Currently, it includes a proposal filter you can use to filter out providers with low reputation scores.
+ *
+ * Reputation data is gathered by the following project: https://github.com/golemfactory/reputation-auditor
+ *
+ * You can adjust the weights used to calculate the score for proposals by using the `setProposalWeights` method.
+ *
+ * NOTE: This class is currently experimental and subject to change.
+ *
+ * NOTE: Only providers from polygon network are being tested, so using this class on testnet will not work.
```
grisha87, line 58: Worse, it will throw an error so we could point this out in the docs explicitly :)

I was running into the same issue with my golem-node-validator and whitelisting solution. There I took the alternative approach

> Query the whitelist with the `paymentNetwork` parameter. If the returned data set is empty, act like an "accept all" filter and allow the proposal to be processed by other filters set by the user.

The reason why I did that is that once upon a time I introduced a bug in my validator that led to an empty whitelist over time. What I mean by that is: if you throw an error to the user when the data set is empty or you have a network issue with reaching that endpoint you essentially take responsibility for the uptime of the user's application. If the reputation endpoint is going to become unavailable, so will the user applications.

Just before we continue - is this the behaviour your're willing to introduce?

examples/experimental/reputation/basic.ts
```diff
@@ -0,0 +1,41 @@
...
+  console.log("The test will start in 5 seconds...");
+  await sleep(5, false);
+
+  const reputation = await ReputationSystem.create({
+    paymentNetwork: "polygon",
+  });
+
+  console.log("Listed providers:", reputation.getData().providers.length);
+
+  const executor = await TaskExecutor.create({
```
grisha87, line 25: FYI: @mgordel

grisha87 (approved):

PR #827, opened 2024-02-29
JST-781: ESM build

SewerynKras:


.github/workflows/ci.yml
mgordel, line 1: we should add this check to `release.yml` in "Perform regular checks"

mgordel (approved):

PR #838, opened 2024-03-06
feat: experimental reputation client agreement selector

pgrzy-golem:
This feature will allow to choose the best available provider based on benchmarking scores.

### DIFF
--- a/src/experimental/reputation/system.ts
+++ b/src/experimental/reputation/system.ts
@@ -74,7 +74,7 @@ export class ReputationSystem {
    * Reputation data.
    */
   private data: ReputationData = {
-    providers: [],
+    testedProviders: [],
   };
 
   /**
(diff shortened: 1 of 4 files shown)


src/experimental/reputation/system.ts
```diff
@@ -41,7 +43,14 @@ export const DEFAULT_AGREEMENT_WEIGHTS: ReputationWeights = {
...
  */
-export const DEFAULT_REPUTATION_URL = "https://reputation.dev-test.golem.network/v1/providers/scores";
+export const DEFAULT_REPUTATION_URL = "https://reputation.dev-test.golem.network/v2/providers/scores";
+
+/**
+ * The number of top scoring providers to consider when selecting an agreement.
+ *
+ * Default for `topPoolSize` agreement selector option.
+ */
+export const DEFAULT_AGREEMENT_TOP_POOL_SIZE = 2;
```
grisha87, line 53: Just curious if that's not to small for the default?

src/experimental/reputation/types.ts
```diff
@@ -25,13 +29,41 @@ export interface ReputationProviderEntry {
   scores: ReputationProviderScores;
 }
 
+/**
+ * Information about a rejected operator.
+ */
+export interface ReputationRejectedOperator {
+  wallet: string;
```
grisha87, line 36: For consistency, please note that we have the `ProviderInfo` type available. It contains `id`, `name` and `WalletAddress`. Perhaps you could use that type here as well.

grisha87 (approved):

src/experimental/reputation/system.ts
pgrzy-golem: It is small, but also the pool of providers is not big. In my tests keeping this value low did affect the results by alot (10-20%).

Due to the limited amount of good proposals readily available, pool size of 5 had pretty much the same result as default random agreement selector.

grisha87 (approved):
Looks good, but the build is failing.

PR #840, opened 2024-03-07
JST-752: upload gvmi directly to provider

SewerynKras:


examples/experimental/local-image/serveLocalGvmi.ts
```diff
@@ -0,0 +1,40 @@
...
+    localImageServer: server,
+  },
+});
+
+async function main() {
+  console.log("Serving local image to the providers...");
+  await server.serve();
+  await golem.init();
+
+  const job = golem.createJob<string>();
```
grisha87, line 22: Not a fan of using the `experimental` APiI here, but I know where you're coming from :)

src/experimental.ts
```diff
@@ -1,3 +1,4 @@
 export * from "./experimental/job";
 export * from "./experimental/golem_network";
 export * from "./experimental/reputation";
+export * from "./experimental/gvmi";
```
grisha87, line 4: I'm curious why you want to have this as an experimental feature? Are there any known problems that prevent it from being an actual feature?

tests/e2e/localImage.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  it("allows the provider to download the image directly from me", async () => {
+    golemClient = new GolemNetwork({});
+    server = serveLocalGvmi(localImagePath);
+    await server.serve();
+    await golemClient.init();
+
+    const job = golemClient.createJob<string>({
+      package: {
+        localImageServer: server,
```
grisha87, line 22: Let me put this in context of TaskExecutor for a moment - I think the target picture looks like this:

```ts
const te = await TaskExecutor.create({
  package: "gftp-serve://./path/to/image.gvmi"
});
```

At the end of the day we don't want them to import the GvmiServer anywhere to use it.

examples/experimental/local-image/serveLocalGvmi.ts
SewerynKras: I think we can remove this example and add one in `task-executor` repo once the feature is released

tests/e2e/localImage.spec.ts
SewerynKras: How will the user get this url? How will they get the image hash? Or do you think the TE should start it's own GVMI server inside and close it once it's finished?

src/experimental.ts
SewerynKras: Hmm now that I think about it there is no good reason, it can be moved to the root export

src/gvmi/gvmiServer.ts
```diff
@@ -0,0 +1,72 @@
...
+ * the server will calculate the hash automatically.
+ */
+export class GvmiServer {
+  private gftp: GftpStorageProvider;
+  public readonly fileHash: string;
+  private fileUrl?: string;
+
+  constructor(private gvmiPath: string) {
+    // no need to check if we're in a node environment here because it's already checked in GftpStorageProvider
+    this.gftp = new GftpStorageProvider(nullLogger());
```
mgordel, line 19: maybe this should be an option in the constructor? If there are any errors while transferring the file (this sometimes happened), we will not see the logs...

examples/local-image/serveLocalGvmi.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  try {
+    const agreement = await agreementPool.getAgreement();
+    payment.acceptPayments(agreement);
+
+    const activity = await Activity.create(agreement, yagna);
+    // Stop listening for new proposals
+    await market.end();
+
+    const ctx = new WorkContext(activity, {});
```
mgordel, line 56: Just wondering, if we had an example where we were uploading files using `ctx.uploadFIle` or `ctx.downloadFile`, we would have to first create `StorageProvider` instance and do `init`, then pass it to the `WorkContext` constructor.. 
and in this case we would have two instances of the `StorageProvider` server. So maybe this `serveLocalGvmi` helper should take `StorageProvider?` as a dependency and only create it when it is null?

mgordel (approved):

PR #868, opened 2024-03-20
Remove EventTarget and replace it with EventEmitters

SewerynKras:


src/activity/factory.ts
```diff
@@ -13,6 +17,7 @@ import { GolemInternalError } from "../error/golem-error";
  */
 export class ActivityFactory {
```
mgordel, line 18: I'm just wondering... how will the user be able to listen to the `activityCreated` event? `ActivityFactory` is not public... it is just a helper for the `Activity.create` function.. so in future low level api, it would be:
```ts
agreement.createActivity();
agreement.events.on('activityCreated');
// underneath, if agreement uses a factory to create activity, it will have to pass this event from the factory to its events

activityPool.events.on('activityCreated'); // or current workService
```
but for current `TaskExecutor` If I want to somehow catch and pass all events from golem-js under the `golemEvents` field, how can I do this?

mgordel (commented):
so I guess in every case when we use `static Entity.create()` we should make it possible to access `Created` events..

mgordel (approved):
We decided that it would remain in this form. When rebuilding a new API, most likely events emitted by factories will be taken over by the objects creating them

PR #870, opened 2024-03-21
fix(payments): fixed payment acceptance for big number amount

mgordel:
Added `amountPrecise` in Invoice and `totalAmountDuePrecise` in DebitNote. The old fields `amount` and `totalAmountDue` are marked as **deprecated**.
The number type caused problems for big numbers. The new `amountPrecise` and `totalAmountDuePrecise` fields are of the string type, and it is recommended to use them instead of the ones mentioned above. Payment events also have new fields for storing `amountPrecise` values.


SewerynKras (approved):
lgtm

PR #876, opened 2024-03-27
feat(activity): added option for maximum number of retries retrieving results when an error occurs

mgordel:


tests/unit/activity.test.ts
```diff
@@ -372,10 +366,41 @@ describe("Activity", () => {
...
+
+      return new Promise<void>((res) => {
+        results.on("error", (error: GolemWorkError) => {
+          expect(error).toBeInstanceOf(GolemWorkError);
+          expect(error.code).toEqual(WorkErrorCode.ActivityResultsFetchingFailed);
+          expect(error.getActivity()).toBeDefined();
+          expect(error.getAgreement()).toBeDefined();
+          expect(error.getProvider()?.name).toEqual("Test Provider");
+          expect(error.previous?.toString()).toEqual(
+            "Error: Failed to fetch activity results. Attempt: 2. Error: timeout",
```
SewerynKras, line 403: I'm not 100% sure what is this test actually testing? Is it the "Attempt: 2" line instead of "Attempt 1"? What do you think about checking how many times `.getExecBatchResults` was called? Or making it so the first time it throws and second time returns results, and we expect results (because it was retried)?

mgordel: you're right, I can check it better

src/activity/activity.ts
```diff
@@ -244,25 +255,25 @@ export class Activity {
...
-              activityId,
-              batchId,
-              undefined,
-              activityExeBatchResultPollIntervalSeconds,
+            const rawExecBachResults = await retry(
+              async (bail, attempt) => {
+                logger.debug(`Trying to poll for batch execution results from yagna. Attempt: ${attempt}`);
+                return api.activity.control
+                  .getExecBatchResults(activityId, batchId, undefined, activityExeBatchResultPollIntervalSeconds)
+                  .catch((err) => handleError(err, bail, attempt));
```
grisha87, line 263: I don't think that the bail method should ever go out to the `handleError` method. The `handleError` can be used in many other contexts where the bail doesn't make sense. Instead, within the `retry` I would call `bail`, this will lead to a throw of the error and then I would pass the error to the `handleError` function.

grisha87 (changes requested):

grisha87 (approved):

PR #882, opened 2024-04-02
Deposit in allocation

SewerynKras:
The deposit details can be added when calling `paymentService.createAllocation` :
```ts
  const allocation = await payment.createAllocation({
    budget: 1.0,
    expirationSec: DURATION_SEC,
    deposit: {
      contract: "0x...",
      id: "0x...",
    }
  });
```

I also added `AllocationOptions` to `PaymentOptions` so it can be added to TE:
```ts
 const executor = await TaskExecutor.create({
    package: "golem/alpine:latest",
    logger: pinoPrettyLogger(),
    allocation: {
      deposit: {
        contract: "0x...",
        id: "0x...",
      },
    },
  });
  ```
  
  related to: JST-829

### DIFF
--- a/src/payment/allocation.ts
+++ b/src/payment/allocation.ts
@@ -13,6 +13,10 @@ export interface AllocationOptions extends BasePaymentOptions {
     platform: string;
   };
   expirationSec?: number;
+  deposit?: {
+    contract: string;
+    id: string;
+  };
 }
 
 /**
@@ -52,6 +56,7 @@ export class Allocation {
         remainingAmount: "",
         spentAmount: "",
         allocationId: "",
+        deposit: config.deposit,
       };
       const newModel = await yagnaApi.payment.createAllocation(model);
       config.logger.debug(
--- a/src/payment/allocation.ts
+++ b/src/payment/allocation.ts
@@ -13,8 +13,20 @@ export interface AllocationOptions extends BasePaymentOptions {
     platform: string;
   };
   expirationSec?: number;
+  /**
+   * Optionally provide a deposit to be used for the allocation, instead of using funds from the yagna wallet.
+   * Deposit is a way to pay for the computation using someone else's funds. Learn more about deposits and allowance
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Review by mgordel — APPROVED** [review]

File: src/payment/allocation.ts [hunk 1]
```diff
@@ -13,6 +13,22 @@ export interface AllocationOptions extends BasePaymentOptions {
     platform: string;
   };
   expirationSec?: number;
+  /**
+   * Optionally provide a deposit to be used for the allocation, instead of using funds from the yagna wallet.
```
grisha87, line 17: The code change seems to be OK, I'd like to suggest changes around the documentation.

This particular field can be just commented on with:

> Instead of using funds from the yagna wallet, you can opt to provide a deposit to be used for allocation.

And then I would suggest creating a standalone interface representing the `Deposit` that will document the `contract` and `id` just as you did it. Then I would document the `Deposit` interface with the explanation _why?_ this is needed. Specifically:

> GLM funds deposit represents a deposit that can be used to settle payments for invoices in the Golem Network.
>
> You have the option to choose deposits instead of the standard allocation when requesting resources from the network. By making use of funds deposited by someone else, such as your application's user, you can avoid paying from your funds. To do this, your end user must deposit the contract. Once the deposit is made, your application can request work and resources on the network, and finally pay the providers from the deposited funds."

If the above holds true, feel free to introduce it to the code documentation :)

grisha87 (approved):

src/payment/allocation.ts
SewerynKras: IMO linking to the official documentation and tutorial is better than writing a paragraph here. `Deposit is a way to pay for the computation using someone else's funds. Learn more about deposits and allowance in the [official documentation]` should be enough to give someone an idea what this does

grisha87: I would agree only if the official docs are in place - while right now they are not, thus the suggestion is to document it as best as we can right here.

SewerynKras:
Closing in favor of https://github.com/golemfactory/golem-js/pull/983

PR #890, opened 2024-04-10
Subscription and batch subscription

SewerynKras:


grisha87:
We decided to go with `rxjs` for now.

PR #897, opened 2024-04-15
feat(market): updated implementation of the proposal pool

grisha87:
JST-868


mgordel (approved):
looks ok to me. There will be some conflicts for the new API version, but we can merge it first.

PR #898, opened 2024-04-15
New API and examples

mgordel:
We should discuss the `ActivityPoolOptions` type, because it is now different for `ActivityPool` and for the builder that uses the `.createActivityPool(name: string, options: ActivityPoolOptions)` method


examples/basic/hello-world.ts
```diff
@@ -1,71 +1,54 @@
...
-    console.log(result);
-
-    // Stop the activity on the provider once you're done
-    await activity.stop();
-    // Release the agreement without plans to re-use -> it's going to terminate this agreement
-    await agreementPool.releaseAgreement(agreement.id, false);
+    const result = await ctx.run("echo Hello World");
+    console.log(result.stdout);
+    proposalSubscription.cancel();
+    await proposalPool.drain();
```
grisha87, line 50: I dropped the `drain` idea, it's not implemented in the `DraftOfferProposalPool`.

examples/deployment/new-api.ts
```diff
@@ -22,48 +22,50 @@ async function main() {
...
-      },
-      replicas: 10,
-      network: "basic",
-      market: {
-        rentHours: 12 /* REQUIRED */,
-        pricing: {
-          maxStartPrice: 1 /* REQUIRED */,
-          maxCpuPerHourPrice: 1 /* REQUIRED */,
-          maxEnvPerHourPrice: 1 /* REQUIRED */,
+      demand: {
```
grisha87, line 25: Yeah, let's discuss this a bit, the second argument shouldn't need that `demand` here. Try to propose different types that will satisfy this requirement.

examples/deployment/new-api.ts
```diff
@@ -22,48 +22,50 @@ async function main() {
...
+          withoutOperators: ["0x123123"],
         },
-        withProviders: ["0x123123"],
-        withoutProviders: ["0x123123"],
-        withOperators: ["0x123123"],
-        withoutOperators: ["0x123123"],
       },
-      payment: {},
-      // dataTransferProtocol: "gftp",
+      pool: { replicas: { min: 1 } },
```
grisha87, line 48: People also shouldn't be defining parameters of the pool. The `replicas` count should be set at the same level as before the change. It appears that we need a dedicated type for this "ComputePowerDemandSpec".

examples/basic/hello-world.ts
mgordel: so should I use `proposalPool.clear()` instead?

mgordel:
As we agreed, I changed the type passed to the builder as `CreateActivityPoolOptions`

examples/basic/hello-world.ts
```diff
@@ -1,71 +1,56 @@
...
+        withoutOperators: ["0x123123"],
+      },
+    };
+    const invoiceFilter = () => true;
+    const debitNoteFilter = () => true;
+    const proposalPool = new DraftOfferProposalPool();
+    const proposalSubscription = await modules.market.startCollectingProposal(demandOptions, proposalPool);
+    const draftProposal = await proposalPool.acquire();
+    const agreement = await modules.market.proposeAgreement(modules.payment, draftProposal, { invoiceFilter });
+    const activity = await modules.activity.createActivity(modules.payment, agreement, { debitNoteFilter });
```
grisha87, line 47: I guess the movement of the debit note filter to agreement creation code will be done in a separate PR?

src/agreement/agreement-pool.ts
```diff
@@ -1,16 +1,16 @@
...
 import { AgreementDTO } from "./service";
 import { EventEmitter } from "eventemitter3";
-import { GolemUserError } from "../shared/error/golem-error";
+import { PaymentModule } from "../payment";
 
 export interface AgreementPoolOptions {
   logger?: Logger;
-  market: MarketOptions;
-  pool?: GenericPoolOptions;
+  poolOptions?: GenericPoolOptions;
```
grisha87, line 12: Just make sure that these "generic pool options" are not leaked out to the user-code, we don't want these to be the dependency of our user.

grisha87 (approved):

examples/basic/hello-world.ts
mgordel: yes

src/agreement/agreement-pool.ts
mgordel: it does not leak from the deployment API level, but if the user wants to use `AgreementPool` directly as in the `basic/hello-world.ts` example, he will be able to set custom parameters for the generic pool there..

PR #900, opened 2024-04-17
Implement market interactions with rxjs

SewerynKras:


mgordel (approved):
I think it can be merged with beta and then I will try to merge the changes with the activity pool.

PR #906, opened 2024-04-19
fix(beta): stabilized examples so that they work, fixed pool implementation

grisha87:
Working with payments still a TODO, addressed in a separate ticket. Fixing pool implementation means: to drain and clear (destroy) all instances upon a single request. Improved error logging because the current implementation was hard to debug and troubleshoot.

### DIFF
--- a/src/activity/activity-pool.ts
+++ b/src/activity/activity-pool.ts
@@ -134,32 +134,29 @@ export class ActivityPool {
     return this.activityPool.available;
   }
 
+  /**
+   * Wait till the pool is ready to use (min number of items in pool are usable)
+   */
+  ready(): Promise<void> {
+    return this.agreementPool.ready();
+  }
+
   private createPoolFactory(): Factory<WorkContext> {
     return {
       create: async (): Promise<WorkContext> => {
-        try {
-          this.logger.debug("Creating new activity to add to pool");
-          const agreement = await this.agreementPool.acquire();
-          const activity = await this.modules.activity.createActivity(agreement);
-          const ctx = new WorkContext(activity, {});
-          await ctx.before();
-          this.events.emit("created", ctx.getDto());
-          return ctx;
-        } catch (err) {
-          this.logger.error("Failed to create new activity for the pool", err);
-          throw err;
-        }
+        this.logger.debug("Creating new activity to add to pool");
+        const agreement = await this.agreementPool.acquire();
+        const activity = await this.modules.activity.createActivity(agreement);
+        const ctx = new WorkContext(activity, {});
+        await ctx.before();
+        this.events.emit("created", ctx.getDto());
+        return ctx;
       },
       destroy: async (ctx: WorkContext) => {
-        try {
-          this.logger.debug("Destroying activity from the pool");
-          await this.modules.activity.destroyActivity(ctx.activity);
-          await this.agreementPool.release(ctx.activity.agreement);
-          this.events.emit("destroyed", ctx.getDto());
-        } catch (err) {
-          this.logger.error("Failed to destroy the activity in the pool", err);
-          throw err;
-        }
+        this.logger.debug("Destroying activity from the pool");
+        await this.modules.activity.destroyActivity(ctx.activity);
+        await this.agreementPool.release(ctx.activity.agreement);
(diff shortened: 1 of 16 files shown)


src/activity/activity-pool.ts
```diff
@@ -127,17 +136,27 @@ export class ActivityPool {
...
-        await this.modules.activity.destroyActivity(activity.activity);
-        await this.agreementPool.release(activity.activity.agreement);
+        try {
+          this.logger.debug("Destroying activity from the pool");
+          await this.modules.activity.destroyActivity(activity.activity);
+          await this.agreementPool.release(activity.activity.agreement);
+        } catch (err) {
+          this.logger.error("Failed to destroy the activity in the pool", err);
+          throw err;
+        }
```
mgordel, line 159: doubled in ` this.activityPool.on("factoryDestroyError"`

src/agreement/agreement-pool.ts
```diff
@@ -108,15 +117,24 @@ export class AgreementPool {
...
-        //TODO: make Agreement compatible with ProposalNew instead of Proposal
-        await this.proposalPool.remove(agreement.proposal);
+        try {
+          this.logger.debug("Destroying agreement from the pool");
+          await this.modules.market.terminateAgreement(agreement, "Finished");
+          await this.proposalPool.remove(agreement.proposal);
+        } catch (err) {
+          this.logger.error("Failed to destroy the agreement in the pool", err);
+          throw err;
+        }
```
mgordel, line 137: handled by `this.agreementPool.on`

src/activity/activity-pool.ts
mgordel, line 1: I wonder if the `.ready()` method would be useful in this interface, which would wait for a situation in which the pool is filled with the indicated number of replicas..
In this implementation we do not know when the pool reaches this state..

```ts
async ready(): Promise<void> {
  return this.activityPool.ready();
}
```

src/activity/activity-pool.ts
```diff
@@ -97,8 +97,17 @@ export class ActivityPool {
...
+   * Sets the pool into draining mode and then clears it
+   *
+   * When set to drain mode, no new acquires will be possible. At the same time, all activities in the pool will be destroyed on the Provider.
+   *
+   * @return Resolves when all activities in the pool are destroyed
+   */
+  async drainAndClear() {
+    await this.activityPool.drain();
+    await this.activityPool.clear();
+    return;
```
mgordel, line 110: we should emit the `end` event here

src/activity/activity-pool.ts
```diff
@@ -97,8 +97,17 @@ export class ActivityPool {
     this.events.emit("destroyed", activity.getDto());
```
mgordel, line 97: this event should be emitted from the function in `createFactoryPool` because it will not be emitted here if the activity is destroyed during the `drainAndClear` process

src/agreement/agreement-pool.ts
```diff
@@ -85,8 +85,17 @@ export class AgreementPool {
     this.events.emit("destroyed", agreement.getDto());
```
mgordel, line 85: this event should be emitted from the function in createFactoryPool because it will not be emitted here if the agreemnet is destroyed during the `drainAndClear` process

src/activity/activity-pool.ts
```diff
@@ -127,17 +136,27 @@ export class ActivityPool {
...
+          this.logger.debug("Creating new activity to add to pool");
+          const agreement = await this.agreementPool.acquire();
+          const activity = await this.modules.activity.createActivity(agreement);
+          const ctx = new WorkContext(activity, {});
+          await ctx.before();
+          return ctx;
+        } catch (err) {
+          this.logger.error("Failed to create new activity for the pool", err);
+          throw err;
+        }
```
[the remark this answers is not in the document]
grisha87: I see what you mean. I need to test this behavior if I will be seeing the activity creation errors.

src/activity/activity-pool.ts
grisha87: + the same for agreement pool

mgordel (approved):

PR #910, opened 2024-04-25
feat(market): added batching initial proposals

mgordel:


src/market/market.module.ts
```diff
@@ -229,7 +251,24 @@ export class MarketModuleImpl implements MarketModule {
         }
         longPoll();
       };
+      // reduce initial proposals to a set grouped by the provider's key to avoid duplicate offers
+      const batch = async () => {
```
SewerynKras, line 255: The implementation is good but I think this should be in a separate operator. I think we want to filter out proposals we don't want first, before putting them in the buffer, plus it's better for readability and maintainability.

I was thinking something like:
```ts
deduplicateProposals(options): OperatorFunction<ProposalNew, ProposalNew> {
  return (source) => 
    new Observable((destination) => {
     const proposalsBatch = ...
     source.subscribe((proposal) => {...});
 })
}

// in `startCollectingProposals`
      switchMap((demand) => this.subscribeForProposals(demand)),
      // for each proposal collected -> filter out undesired and invalid ones
      filter((proposal) => proposal.isValid()),
      filter((proposal) => !options.filter || options.filter(proposal)),
      this.deduplicateProposals(options),   //<---------------
      ...
```

SewerynKras (changes requested):

src/market/market.module.ts
mgordel: It has been moved to a separate function

SewerynKras (approved):

PR #911, opened 2024-04-25
Market API adapter and deprecate Allocation.create and Package

SewerynKras:
This PR cannot be merged at the moment, I'm waiting for https://github.com/golemfactory/golem-js/pull/909 to be error-free before rebasing it here and fixing merge conflicts

### DIFF
--- a/examples/basic/hello-world.ts
+++ b/examples/basic/hello-world.ts
@@ -1,4 +1,4 @@
-import { Allocation, DraftOfferProposalPool, GolemNetwork, Package } from "@golem-sdk/golem-js";
+import { DraftOfferProposalPool, GolemNetwork } from "@golem-sdk/golem-js";
 
 import { pinoPrettyLogger } from "@golem-sdk/pino-logger";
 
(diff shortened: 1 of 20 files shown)


examples/basic/hello-world.ts
```diff
@@ -41,27 +41,13 @@ import { pinoPrettyLogger } from "@golem-sdk/pino-logger";
       logger,
     });
 
-    const workload = Package.create({
-      imageTag: demand.demand.image,
+    const allocation = await glm.payment.createAllocation({ budget: 1 });
+    const demandSpecification = await glm.market.buildDemand(demand.demand, allocation);
```
mgordel, line 45: I don't think we will need allocation at all here, there was an idea that decorations from allocation could be generated manually based on the requestor id..

src/market/builder.ts
```diff
@@ -53,6 +54,9 @@ export class DecorationsBuilder {
       constraints: this.constraints.map((c) => `(${c.key + c.comparisonOperator + c.value})`),
     };
   }
+  /**
+   * @deprecated
```
mgordel, line 58: ? this is used below in the new getDemandSpecification method.. ?

mgordel (approved):

PR #915, opened 2024-04-29
JST-874: divorce allocation from demand

SewerynKras:


src/payment/payment.module.ts
```diff
@@ -42,6 +42,11 @@ export type PaymentModuleConfig = {
   logger?: Logger;
 };
 
+export interface PayerDetails {
+  address: string;
+  platform: string;
```
grisha87, line 47: It looks like we could model `address`, `platform`, `driver` under the same hood? Just a moment ago we had this class - `PaymentPlatform`. I can see it's in the codebase but it's not used. Perhaps we can bundle all of the "payer/payee" information under a single type.

grisha87 (approved):

src/payment/payment.module.ts
SewerynKras: Good idea, I changed the name to `PayerDetails` so it doesnt clash with what payment platform actually is

PR #921, opened 2024-05-08
refactor(market): remove old Demand class and leave the new implementation

grisha87:
First iteration of the changes for review and further modifications.

BREAKING CHANGE: This refactor removes several deprecated types and modules from the market module in an effort to enhance code readability and maintainability. This includes the deletion of classes such as Package, DemandConfig, and various DemandDirectorConfig classes.


src/market/demand.ts
mgordel, line 1: If we have a `demand` folder, maybe this file should be there?

src/market/market.module.ts
```diff
@@ -189,49 +205,41 @@ export class MarketModuleImpl implements MarketModule {
...
-    return builder.getDemandSpecification(payerDetails.getPaymentPlatform(), demandSpecificConfig.expirationSec);
+  async buildDemandDetails(options: BuildDemandOptions, payerDetails: PayerDetails): Promise<DemandSpecification> {
+    const builder = new DemandBodyBuilder();
+
+    // Instruct the builder what's required
+    const basicConfig = new BasicDemandDirectorConfig(options.basic);
+    const basicDirector = new BasicDemandDirector(basicConfig);
+    basicDirector.apply(builder);
+
+    const workloadOptions = options.activity
```
mgordel, line 216: I am considering the name... so we use the name workload here, whereas in the options we use the name activity... I think I like the name `workload` better than `activity` in this context.. but this is just a suggestion

mgordel (approved):
It reads nicely, I didn't notice any gaps

PR #922, opened 2024-05-08
JST-883: payment in pools

SewerynKras:


src/agreement/lease-process-pool.ts
```diff
@@ -0,0 +1,285 @@
...
+
+  /**
+   * Wait till the pool is ready to use (min number of items in pool are usable).
+   * Optionally, you can pass an AbortSignal to cancel the operation.
+   */
+  async ready(abortSignal?: AbortSignal): Promise<void> {
+    if (this.minPoolSize <= this.getAvailable()) {
+      return;
+    }
+    const signal = abortSignal || AbortSignal.timeout(10_000);
```
grisha87, line 265: `AbortSignal.timeout` -> nice one! I learned something today!
Nevertheless, I don't think that's good to assume this for the user. I would suggest:

- not having this set up by default
- user can opt-in by providing a corresponding option - `readyTimeoutSec` for example

src/agreement/lease-process-pool.ts
```diff
@@ -0,0 +1,285 @@
...
+    this.borrowed.add(leaseProcess);
+    this.events.emit("acquired", leaseProcess.agreement.getDto());
+    return leaseProcess;
+  }
+
+  /**
+   * If there are any acquires waiting in the queue, the lease process will be passed to the first one.
+   * Otherwise, the lease process will be added to the queue.
+   */
+  private addLeaseProcessToQueue(leaseProcess: LeaseProcess) {
```
grisha87, line 182: Suggested refactoring so that tired developers can parse it quicker :)

```ts
private passToNextAcquireOrKeep(leaseProcess: LeaseProcess)
```

src/agreement/lease-process-pool.test.ts
```diff
@@ -0,0 +1,344 @@
...
+        .mockResolvedValue({ agreement: { id: "1", getDto: () => ({}) } } as LeaseProcess);
+
+      const pool = getLeasePool({ min: 3 });
+      expect(pool.getSize()).toBe(0);
+      await pool.acquire();
+      expect(pool.getSize()).toBe(1);
+      expect(pool.getBorrowed()).toBe(1);
+      expect(pool.getAvailable()).toBe(0);
+    });
+    it("waits for a lease to become available when the pool is full", async () => {
```
grisha87, line 146: What was really neat about the generic-pool was that this waiting time could be also controlled with a timeout. Will we need `acquireTimeoutSec` in golem-js? Considering the use case from TaskExecutor: I have 100 tasks to run and only 5 workers (lease processes). I think it would be beneficial to be able to specify that the acquire timeout to short-circuit the waiting time. Users could specify `acquire({ timeoutSec: 15 })` to fail the operation if they didn't get the resource needed.

---

This reminds me about another feature that the generic-pool had: the eviction mechanism - if a resource was sitting idle for a configured amount of time, it was destroyed. It was a nice auto-scaling feature of generic-pool.

---

I wouldn't like to bloat this PR with additional requirements. I'm going to put them in our backlog to implement next.

grisha87 (changes requested):

src/agreement/lease-process-pool.ts
```diff
@@ -0,0 +1,285 @@
...
+  }
+
+  /**
+   * Take the first valid lease process from the pool
+   * If there is no valid lease process, return null
+   */
+  private async takeValidLeaseProcess(): Promise<LeaseProcess | null> {
+    let leaseProcess: LeaseProcess | null = null;
+    if (this.highPriority.size > 0) {
+      leaseProcess = this.highPriority.values().next().value as LeaseProcess;
```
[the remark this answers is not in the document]
SewerynKras: Surprisingly the correct way to write this in ts is `[leaseProcess] = this.highPriority.keys()` but I think the version we have now is more readable

src/agreement/lease-process-pool.ts
```diff
@@ -0,0 +1,285 @@
...
+    this.isDraining = true;
+    this.acquireQueue = [];
+    const allLeaseProcesses = Array.from(this.borrowed)
+      .concat(Array.from(this.lowPriority))
+      .concat(Array.from(this.highPriority));
+    await Promise.allSettled(allLeaseProcesses.map((leaseProcess) => this.destroy(leaseProcess)));
+    this.lowPriority.clear();
+    this.highPriority.clear();
+    this.borrowed.clear();
+    this.isDraining = false;
```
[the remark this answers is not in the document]
SewerynKras: I'm treating `drainAndClear` like `reset`, if someone wants to reuse the same pool instance later then why stop them

src/agreement/lease-process-pool.ts
SewerynKras: 1. I think there needs to be some sort of short-circuit when we're dealing with a potentially infinite loop like this one, otherwise this will hang forever (or until ctrl-c)
2. If we use `AbortSignal` then we have both options - a timeout with `AbortSignal.timeout` or manual short-circuit with `AbortController`. Using `readyTimeoutSec: number` gives us only the first option with seemingly no benefit?

src/agreement/lease-process-pool.ts
```diff
@@ -0,0 +1,285 @@
...
+  /**
+   * Borrow a lease process from the pool. If there is no valid lease process a new one will be created.
+   */
+  async acquire(): Promise<LeaseProcess> {
+    if (this.isDraining) {
+      throw new Error("The pool is in draining mode");
+    }
+    let leaseProcess = await this.takeValidLeaseProcess();
+    if (!leaseProcess) {
+      if (!this.canCreateMoreLeaseProcesses()) {
```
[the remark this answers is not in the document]
SewerynKras: `enqueueAcquire` could be so much cleaner if we dropped support for node<22 :smiling_face_with_tear: 
```ts
    const {resolve, promise} = Promise.withResolvers<LeaseProcess>();
    this.acquireQueue.push((leaseProcess) => {
      this.borrowed.add(leaseProcess);
      this.events.emit("acquired", leaseProcess.agreement.getDto());
      resolve(leaseProcess);
    });
    return promise;
```

PR #927, opened 2024-05-15
Network module

mgordel:


src/network/network.module.ts
```diff
@@ -1,11 +1,213 @@
...
+        if (!network.isNodeIpUnique(ipv4)) {
+          throw new GolemNetworkError(
+            `IP '${nodeIp.toString()}' has already been assigned in this network.`,
+            NetworkErrorCode.AddressAlreadyAssigned,
+            network.getNetworkInfo(),
+          );
+        }
+      } else {
+        ipv4 = network.getFirstAvailableIpAddress();
+      }
```
SewerynKras, line 165: This is really hard to read with all that nesting, what do you think about a refactor to move the logic of finding a free ip to a separate function?
```ts
private getFreeIpInNetwork(network: Network, targetIp?: string): IPv4 {
    if (!targetIp) {
      return network.getFirstAvailableIpAddress();
    }
    const ipv4 = IPv4.fromString(targetIp);
    if (!network.isIpInNetwork(ipv4)) {
      throw new GolemNetworkError(
        `The given IP ('${targetIp}') address must belong to the network ('${network.getNetworkInfo().ip}').`,
        NetworkErrorCode.AddressOutOfRange,
        network.getNetworkInfo(),
      );
    }
    if (!network.isNodeIpUnique(ipv4)) {
      throw new GolemNetworkError(
        `IP '${targetIp.toString()}' has already been assigned in this network.`,
        NetworkErrorCode.AddressAlreadyAssigned,
        network.getNetworkInfo(),
      );
    }
    return ipv4;
  }

  async createNetworkNode(network: Network, nodeId: string, nodeIp?: string): Promise<NetworkNode> {
    if (!network.isNodeIdUnique(nodeId)) {
      throw new GolemNetworkError(
        `Network ID '${nodeId}' has already been assigned in this network.`,
        NetworkErrorCode.AddressAlreadyAssigned,
        network.getNetworkInfo(),
      );
    }
    try {
      const ipv4 = this.getFreeIpInNetwork(network, nodeIp);
      const node = await this.deps.networkApi.createNetworkNode(network, nodeId, ipv4.toString());
      network.addNode(node);
      this.deps.logger.info(`Node has been added to the network.`, { id: nodeId, ip: ipv4.toString() });
      return node;
      //.. rest
```

src/network/network.module.ts
```diff
@@ -1,11 +1,213 @@
...
+      const node = await this.deps.networkApi.createNetworkNode(network, nodeId, ipv4.toString());
+      network.addNode(node);
+      this.deps.logger.info(`Node has been added to the network.`, { id: nodeId, ip: ipv4.toString() });
+      return node;
+    } catch (error) {
+      if (error instanceof GolemNetworkError) {
+        throw error;
+      }
+      throw new GolemNetworkError(
+        `Unable to add node to network. ${error?.data?.message || error.toString()}`,
```
SewerynKras, line 175: you can get a better error message than `error?.data?.message || error.toString()` with `getMessageFromApiError`

src/network/network.module.ts
```diff
@@ -1,11 +1,213 @@
...
+      this.deps.logger.info(`Node has been removed from the network.`, {
+        network: network.getNetworkInfo().ip,
+        nodeIp: node.ip,
+      });
+    } catch (error) {
+      if (error instanceof GolemNetworkError) {
+        throw error;
+      }
+      throw new GolemNetworkError(
+        `Unable to remove network node. ${error}`,
```
SewerynKras, line 202: Same as above, `${error}` will not display the response from yagna in case of api error unless you use `getMessageFromApiError`

src/shared/yagna/adapters/network-api-adapter.ts
```diff
@@ -0,0 +1,37 @@
...
+    await this.yagnaApi.net.addNode(network.id, { id: nodeId, ip: nodeIp });
+    return new NetworkNode(nodeId, nodeIp, network.getNetworkInfo.bind(network));
+  }
+  async removeNetworkNode(network: Network, node: NetworkNode): Promise<void> {
+    await this.yagnaApi.net.removeNode(network.id, node.id);
+  }
+
+  async getIdentity() {
+    return this.yagnaApi.identity.getIdentity().then((res) => res.identity);
+  }
```
SewerynKras, line 30: I think calls to yagna api should be wrapped in try-catch blocks and mapped to our domain errors. At least that's what I did in all other adapters

SewerynKras (changes requested):

SewerynKras:
I ran the example a couple times and twice I got this error:
image
I think there is a race condition between the callback that happens in `oneOf`:
```ts
    // Attach handlers for cleanup after all work's done
    lease.events.once("finalized", async () => {
      if (options?.network && networkNode) {
        await this.network.removeNetworkNode(options.network, networkNode);
      }
      await this.payment.releaseAllocation(allocation);
    });
```
and the network being destroyed with `glm.destroyNetwork`.

SewerynKras:
If I add artificial delay to the request that removes the node
(fragment of `network.module.ts`)
```ts
 async removeNetworkNode(network: Network, node: NetworkNode): Promise<void> {
    if (!network.hasNode(node)) {
      throw new GolemNetworkError(
        `The network node ${node.id} does not belong to the network`,
        NetworkErrorCode.NodeRemovalFailed,
        network.getNetworkInfo(),
      );
    }
    try {
      await sleep(1, false); // <-----------------------------------
      await this.deps.networkApi.removeNetworkNode(network, node);
      network.removeNode(node);
```
then I get an error 100% of the time. Maybe we should add an async lock to network operations? (add node, remove node, remove network)

mgordel:
Yes, async lock will be necessary, I was also thinking of making the network state, after remove it would change to `removed` and then for this state node removal would be ignored or only logged with warning

src/network/network.ts
```diff
@@ -73,12 +86,26 @@ export class Network {
...
+        this.getNetworkInfo(),
+      );
+    }
     if (!this.hasNode(node)) {
       throw new GolemNetworkError(`There is no node ${node.id} in the network`, NetworkErrorCode.NodeRemovalFailed);
     }
     this.nodes.delete(node.id);
   }
 
+  public remove() {
```
SewerynKras, line 102: `remove` implies that this does something (like call yagna) but it only sets the state as removed. Maybe we could rename it to something like `markAsRemoved`?

SewerynKras (approved):

SewerynKras:
Implementation looks good, if you have the time you can rewrite the example to use `manyOf` instead of creating two demands

PR #928, opened 2024-05-16
Grisha87/deprecated payment service removal

grisha87:


examples/advanced/event-driven-payments.ts
SewerynKras, line 1: What's the point of this example? If no work is being done then no debit notes will be emitted

examples/basic/hello-world.ts
```diff
@@ -65,8 +74,8 @@ import { pinoPrettyLogger } from "@golem-sdk/pino-logger";
     await proposalPool.remove(draftProposal);
 
     // This will keep the script waiting for payments etc
+    // Also it will release the allocation internally
```
SewerynKras, line 77: why? Allocations are reused across leases, so if we terminate one all others will no longer be able to pay their invoices?

src/deployment/deployment.ts
```diff
@@ -153,7 +153,8 @@ export class Deployment {
     // TODO: allocation is not used in this api?
     // eslint-disable-next-line @typescript-eslint/no-unused-vars
     const allocation = await this.modules.payment.createAllocation({
-      budget: 1,
+      budget: 1, // TODO: Shouldn't this be generated by the deployment internally when the demand is known?
```
SewerynKras, line 156: `this.modules.market.estimateBudget(demand)`

src/deployment/deployment.ts
```diff
@@ -153,7 +153,8 @@ export class Deployment {
...
+      expirationSec: 24 * 60 * 60,
```
SewerynKras, line 157: not a fan of hard-coding this, can't we do `demand.market.rentHours * 60 * 60`?

src/experimental/new-api/index.ts
SewerynKras, line 1: empty file

src/golem-network.ts
```diff
@@ -251,9 +259,13 @@ export class GolemNetwork {
     const agreement = await this.market.proposeAgreement(draftProposal);
     proposalSubscription.unsubscribe();
 
-    const allocation = await this.payment.createAllocation({ budget: 1 });
     const lease = this.market.createLease(agreement, allocation);
 
+    // Attach handlers for cleanup after all work's done
+    lease.events.once("finalized", async () => {
+      await this.payment.releaseAllocation(allocation);
+    });
```
SewerynKras, line 267: If the lease fails to finalize (for example provider never sends the invoice or there is some error while killing activity) then theoretically the allocation will never be cleared :/ Not sure what to do about this tbh, maybe `finalized` should emit even if the function errors

src/shared/utils/waitForCondition.ts
```diff
@@ -12,16 +12,15 @@ import { GolemTimeoutError } from "../error/golem-error";
...
+  check: () => boolean | Promise<boolean>,
   opts = { timeoutSeconds: 30, intervalSeconds: 1 },
 ): Promise<void> {
   let verifyInterval: NodeJS.Timeout | undefined;
   let waitTimeout: NodeJS.Timeout | undefined;
 
   const verify = new Promise<void>((resolve) => {
-    verifyInterval = setInterval(() => {
-      if (check()) {
-        clearInterval(verifyInterval);
```
SewerynKras, line 24: why are we no longer clearing the interval? Wouldnt that call `check` indefinitely?

SewerynKras (changes requested):

examples/advanced/event-driven-payments.ts
grisha87: Merely a draft for an example. I'm going to drop it from the PR. I'm curious about your opinion on this API. The idea was that we could also have examples that can show how people can use rxjs + events with the SDK to build a hello world, but I didn't start implementing the APIs for the other components (market, activity) etc.

examples/basic/hello-world.ts
grisha87: You're right. That's a leftover after one of the changes. I dropped the comment and added `await glm.payment.releaseAllocation(allocation);` below `finalize` call in the example.

src/experimental/new-api/index.ts
grisha87: Dropped

src/golem-network.ts
grisha87: I will cover this and trigger the event in `finally`

src/deployment/deployment.ts
grisha87: It actually could be entirely dropped (this line), as the allocation is not used in that file at any case.

src/deployment/deployment.ts
grisha87: It actually could be entirely dropped (this line), as the allocation is not used in that file at any case.

src/agreement/lease-process.ts
```diff
@@ -75,14 +75,19 @@ export class LeaseProcess {
...
+    try {
+      this.logger.debug("Waiting for payment process of agreement to finish", { agreementId: this.agreement.id });
+      if (this.currentActivity) {
+        await this.activityApi.destroyActivity(this.currentActivity);
+        await this.agreementApi.terminateAgreement(this.agreement);
+      }
+      await waitForCondition(() => this.paymentProcess.isFinished());
+      this.logger.debug("Payment process for agreement finalized", { agreementId: this.agreement.id });
+    } catch (error) {
+      this.logger.error("Payment process finalization failed", { agreementId: this.agreement.id, error });
```
SewerynKras, line 87: we should re-throw the error if finalization fails

src/deployment/deployment.ts
```diff
@@ -150,12 +150,6 @@ export class Deployment {
 
     await this.dataTransferProtocol.init();
 
-    // TODO: allocation is not used in this api?
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    const allocation = await this.modules.payment.createAllocation({
-      budget: 1, // TODO: Shouldn't this be generated by the deployment internally when the demand is known?
-      expirationSec: 24 * 60 * 60,
-    });
```
SewerynKras, line 157: Yeah, it's not used now because agreement pool doesn't take an allocation, but in the parallel universe where we have `leaseProcessPool` it uses this allocation, please leave it here and if possible remove the hard coded budget and expirationSec

examples/advanced/event-driven-payments.ts
SewerynKras: So an even lower level API that doesn't even use lease process but "raw" agreements and activities? Could be useful but not for 3.0

SewerynKras (changes requested):

src/deployment/deployment.ts
grisha87: Addressing this type of allocation within the context of the deployment isn't going to be fast. I'd like to discuss my findings from the implementation attempt I just made, so that we can design how this should look in it's ultimate form - instead of a hack.

SewerynKras (approved):

PR #934, opened 2024-05-22
Many-of

SewerynKras:


src/golem-network/golem-network.ts
```diff
@@ -0,0 +1,379 @@
...
+  WebSocketBrowserStorageProvider,
+} from "../shared/storage";
+
+export interface GolemNetworkOptions {
+  logger?: Logger;
+  api?: {
+    key?: string;
+    url?: string;
+  };
+  market?: Partial<MarketOptions>;
```
mgordel, line 43: this was created earlier, but as we are here it could be improved.
this seems to be unused anywhere, as is `deployment`... only `PaymentModule` has its options, which in addition seem to duplicate `PaymentSpec` in `DemandSpec`.... this needs to be clarified

mgordel (approved):
I tested the example and everything seems ok

PR #936, opened 2024-05-23
Module restructuring

grisha87:
Introducing changes that align high-level, low-level and experimental modules according to what we agreed to during today's call.


src/golem-network.ts
```diff
@@ -1,4 +1,4 @@
-import { DataTransferProtocol, DeploymentOptions, GolemDeploymentBuilder } from "./deployment";
+import { DataTransferProtocol, DeploymentOptions, GolemDeploymentBuilder } from "./experimental/deployment";
```
SewerynKras, line 1: If deployment is moved to experimental then there should be a comment on `GolemNetwork.creteDeploymentBuilder` that it's experimental, or that method should be removed altogether

src/experimental.ts
SewerynKras, line 1: Experimental modules are now exported in two places, in `src/experimental.ts` and in `src/experimental/index.ts`. 
We should either:
-  delete `src/experimental/index.ts` or
-  replace this file with `export * from './experimental'` or
-  delete this file and change rollup config to use `src/experimental/index.ts`

src/golem-network.ts
SewerynKras: Pull changes from beta before doing this because after moving this file to `src/golem-network/` git thinks it's a new file

SewerynKras (changes requested):

src/experimental.ts
grisha87: I'm dropping `src/experimental.ts` and adjusting rollup configs.

SewerynKras (approved):

PR #942, opened 2024-05-27
Invoie and debit note filtering

SewerynKras:


src/market/agreement/service.ts
mgordel, line 1: I don't think it's used anywhere anymore and we can get rid of it

mgordel (approved):

src/market/agreement/service.ts
SewerynKras: Reputation relies on AgreementSelector and candidate, so I moved those to a separate file and deleted agreement pool service

PR #943, opened 2024-05-28
Lease Module

SewerynKras:


src/lease-process/lease.module.ts
```diff
@@ -0,0 +1,78 @@
...
+      paymentProcess,
+      this.deps.marketModule,
+      this.deps.activityModule,
+      this.deps.logger.child("lease-process"),
+      options,
+    );
+    paymentProcess.start();
+    lease.events.once("finalized", () => {
+      paymentProcess.stop();
+    });
```
mgordel, line 57: I just wonder if it is necessary to explicitly start and stop paymentProcess ... ? The PaymentProcess could automatically subscribe in constructor and unsubscribe is finished - when this.isFinished() === true

mgordel (approved):

PR #947, opened 2024-05-29
fix: ability to load env variables for subnet tag and payment network

mgordel:


src/payment/payment.module.ts
```diff
@@ -81,7 +82,7 @@ export class PaymentModuleImpl implements PaymentModule {
 
   private readonly options: Required<PaymentModuleOptions> = {
     driver: "erc20",
-    network: "holesky",
+    network: EnvUtils.getPaymentNetwork() || "holesky",
```
SewerynKras, line 85: `EnvUtils.getPaymentNetwork` and `EnvUtils.getYagnaSubnet` already have defaults if env is empty, no need for the `|| "holesky"` and `|| "public"`

SewerynKras: It's better to have defaults in one place

SewerynKras (changes requested):

SewerynKras (approved):

PR #948, opened 2024-05-29
Fix TODOs left in the codebase

SewerynKras:


src/shared/yagna/adapters/agreement-api-adapter.ts
```diff
@@ -119,7 +111,6 @@ export class AgreementApiAdapter implements IAgreementApi {
       }
 
       await withTimeout(
-        // TODO: Make a fix in ya-ts-client typings so that's going to be specifically {message:string}
```
SewerynKras, line 122: @mgordel @grisha87 Here the spec defines `Reason` as 
```yaml
    Reason:
      type: object
      description: >
        Generic Event reason information structure.
      required:
        - message
      properties:
        message:
          type: string
      additionalProperties: true
```
notice the `additionalProperties: true`, which gets converted to `Record<string, any` :( 
This is a bug in the generator not in ya-client. Apparently this got fixed in the actively maintained fork: https://github.com/ferdikoomen/openapi-typescript-codegen/issues/1909#issuecomment-1925791547

mgordel (approved):

PR #949, opened 2024-05-29
fix(activity): fixed streaming batch via EventSource and runAndStrem example

mgordel:


examples/basic/run-and-stream.ts
```diff
@@ -0,0 +1,48 @@
...
+      level: "info",
+    }),
+  });
+
+  try {
+    await glm.connect();
+    const lease = await glm.oneOf(order);
+    const exe = await lease.getExeUnit();
+
+    const remoteProcess = await exe.runAndStream("sleep 1 && echo 'Hello World' && echo 'Hello Golem' >&2");
```
SewerynKras, line 37: IMO it would be cool if we showcased a command that actually takes some time to complete here so the user can see the messages as they come, maybe even something as simple as:
```ts
    const remoteProcess = await exe.runAndStream(
      `
      sleep 1
      echo -n 'Hello from stdout' >&1
      echo -n 'Hello from stderr' >&2
      sleep 1
      echo -n 'Hello from stdout again' >&1
      echo -n 'Hello from stderr again' >&2
      sleep 1
      echo -n 'Hello from stdout yet again' >&1
      echo -n 'Hello from stderr yet again' >&2
      `,
    );
```

SewerynKras (approved):

examples/basic/run-and-stream.ts
mgordel: you're right, it's better

PR #956, opened 2024-06-04
Payment events

SewerynKras:


src/payment/api.ts
```diff
@@ -3,6 +3,30 @@ import { Invoice } from "./invoice";
 import { DebitNote } from "./debit_note";
 import { Allocation } from "./allocation";
 
+export type PaymentEvents = {
+  allocationCreated: (allocation: Allocation) => void;
+  errorCreatingAllocation: (error: Error) => void;
```
mgordel, line 8: I wonder if, from a user perspective, wouldn't it be better to listen on `glm.payment.on('error', (error: GolemPaymentError))`? instead of specific errors (like `errorCreatingAllocation`) etc.. ?

SewerynKras: `GolemPaymentError` doesn't have all the context, for example if the error is `errorAcceptingInvoice` it's nice to see what invoice we tried to reject, but `GolemPaymentError` only has `allocation` and `provider`. Adding all possible entities to the error makes it super bloated and hard o work with imo

mgordel: yes, both solutions are a bit problematic. ok, let's leave it as it is

mgordel (approved):

PR #957, opened 2024-06-05
Activity events

SewerynKras:


src/activity/activity.ts
```diff
@@ -1,8 +1,5 @@
-import { Logger } from "../shared/utils";
 import { Agreement, ProviderInfo } from "../market/agreement";
 import { EventEmitter } from "eventemitter3";
-import { ExecutionOptions, ExeScriptExecutor } from "./exe-script-executor";
-import { IActivityApi } from "./types";
 
 /** @deprecated */
```
mgordel, line 4: this can already be removed

mgordel (changes requested):

src/activity/activity.ts
SewerynKras: already removed in other PR: https://github.com/golemfactory/golem-js/pull/958/commits/4a7659ece0c271f8384faefb11be41865fc4b4b7

src/activity/activity.module.ts
```diff
@@ -78,52 +108,117 @@ export class ActivityModuleImpl implements ActivityModule {
...
+      throw error;
+    }
+  }
 
-    return updated;
+  async fetchActivity(activityId: Activity["id"]): Promise<Activity> {
+    this.logger.info("Fetching activity state", { activityId });
+    try {
+      const upToDateActivity = await this.activityApi.getActivity(activityId);
+      this.events.emit("activityStateChanged", upToDateActivity, upToDateActivity.getState());
```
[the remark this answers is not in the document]
SewerynKras: activity repo doesn't keep track of known activities because we can get them from yagna :/ also I don't want to jam an event emitter there. I ended up splitting the `fetchActivity` method into two - refresh and find - and emitting the event in `refresh` when the state changes. Let me know if this makes sense or we can have a call tomorrow

src/activity/activity.module.ts
```diff
@@ -78,52 +114,129 @@ export class ActivityModuleImpl implements ActivityModule {
...
+      this.events.emit("activityDestroyed", updated);
+      return updated;
+    } catch (error) {
+      this.events.emit("errorDestroyingActivity", activity, error);
+      throw error;
+    }
+  }
+
+  async refreshActivity(staleActivity: Activity): Promise<Activity> {
+    this.logger.info("Fetching latest activity state", { activityId: staleActivity.id });
```
mgordel, line 201: I think this could be the debug level and additionally include state info

mgordel (approved):

PR #960, opened 2024-06-06
fix(market): added offer pricing filter

mgordel:


src/market/market.module.ts
```diff
@@ -625,4 +611,33 @@ export class MarketModuleImpl implements MarketModule {
...
+  }
+
+  private filterProposalsByPricingOptions(pricing: PricingOptions, proposal: OfferProposal) {
+    let isPriceValid = true;
+    if (pricing.model === "linear") {
+      isPriceValid =
+        proposal.pricing.cpuSec <= pricing.maxCpuPerHourPrice / 3600 &&
+        proposal.pricing.envSec <= pricing.maxEnvPerHourPrice / 3600 &&
+        proposal.pricing.start <= pricing.maxStartPrice;
+    }
```
SewerynKras, line 637: what about `burn-rate`? We can check if start price + envSec * 3600 + cpuSec * 3600 is less than avgGlmPerHour

SewerynKras (changes requested):

SewerynKras (approved):

PR #967, opened 2024-06-07
Update README + add module override example

SewerynKras:


mgordel (approved):
Yes.. we didn't think about the fact that overwritten implementations have a lot of dependencies to pass and it would be bad to require the user to provide them themselves... so I don't see a better solution than your proposed factory implementation..
From the perspective of the example, it looks good.

PR #971, opened 2024-06-10
fix(lease-process): fixed finalizing already terminated agreements

mgordel:
If the agreement is terminated on the provider's side, we should not send a request to yagna for termination.

In addition, in the case of debitNote/Invoice rejection e.g. due to filters, we should not send a request to the endpoint reject - Yagna has not implemented this and always responds 501:NotImplented

### DIFF
--- a/src/payment/agreement_payment_process.ts
+++ b/src/payment/agreement_payment_process.ts
@@ -187,14 +187,17 @@ export class AgreementPaymentProcess {
       await this.paymentModule.rejectDebitNote(debitNote, rejectMessage);
       this.logger.warn(`DebitNote rejected`, { reason: rejectMessage });
     } catch (error) {
-      const message = getMessageFromApiError(error);
-      throw new GolemPaymentError(
-        `Unable to reject debit note ${debitNote.id}. ${message}`,
-        PaymentErrorCode.DebitNoteRejectionFailed,
-        undefined,
-        debitNote.provider,
-        error,
-      );
+      this.logger.warn(`DebitNote rejected`, { reason: rejectMessage });
+      // TODO: this endpoint is not implemented in Yagna, it always responds 501:NotImplemented.
+      // Until it is implemented ny Yagna, it only logs as a warning
+      // const message = getMessageFromApiError(error);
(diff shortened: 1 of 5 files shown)


src/payment/agreement_payment_process.ts
```diff
@@ -282,14 +285,17 @@ export class AgreementPaymentProcess {
...
-      const message = getMessageFromApiError(error);
-      throw new GolemPaymentError(
-        `Unable to reject invoice ${invoice.id} ${message}`,
-        PaymentErrorCode.InvoiceRejectionFailed,
-        undefined,
-        invoice.provider,
-        error,
-      );
+      this.logger.warn(`Invoice rejected`, { reason: message });
+      // TODO: this endpoint is not implemented in Yagna, it always responds 501:NotImplemented.
```
SewerynKras, line 289: I think instead of here, we should comment out the actual call to yagna in the api adapter and move the warning to `payment.module.rejectInvoice`. Just so the behavior is the same no if it's the payment process trying to reject or the user manually

src/payment/agreement_payment_process.ts
```diff
@@ -187,14 +187,17 @@ export class AgreementPaymentProcess {
...
-      const message = getMessageFromApiError(error);
-      throw new GolemPaymentError(
-        `Unable to reject debit note ${debitNote.id}. ${message}`,
-        PaymentErrorCode.DebitNoteRejectionFailed,
-        undefined,
-        debitNote.provider,
-        error,
-      );
+      this.logger.warn(`DebitNote rejected`, { reason: rejectMessage });
+      // TODO: this endpoint is not implemented in Yagna, it always responds 501:NotImplemented.
```
SewerynKras, line 191: same as invoice^

SewerynKras (changes requested):

src/payment/agreement_payment_process.ts
mgordel: It turns out that this is not implemented only for debitNote, I moved the warn to the module and commented out the call to the api

SewerynKras (approved):

PR #974, opened 2024-06-13
fix(lease-process-pool): fixed asynchronous acquiring of lease processes

mgordel:


SewerynKras:
The issue with async-lock is that a) we lose the ability to create leases in parallel and b) only one acquire can wait in `acquireQueue` (because async-lock will not unlock until the promise resolves). I wrote a unit test to test it some more:
```ts
it("doesn't create more processes than allowed", async () => {
      jest.useFakeTimers();
      const pool = getLeasePool({ min: 3, max: 3 });
      pool["createNewLeaseProcess"] = jest.fn(
        () => new Promise((resolve) => setTimeout(() => resolve(getMockLeaseProcess()), 50)),
      );
      expect(pool.getSize()).toBe(0);
      pool.acquire(); // should be resolved after 50ms
      pool.acquire(); // should be resolved after 50ms
      pool.acquire(); // should be resolved after 50ms
      pool.acquire(); // should be added to the queue
      pool.acquire(); // should be added to the queue
      pool.acquire(); // should be added to the queue
      pool.acquire(); // should be added to the queue
      await jest.advanceTimersByTimeAsync(50);
      expect(pool.getSize()).toBe(3);
      expect(pool.getBorrowedSize()).toBe(3);
      expect(pool.getAvailableSize()).toBe(0);
      expect(pool["acquireQueue"].length).toBe(4);
});
```
without async-lock `pool.getSIze() === 7` but with async-lock `pool.getSize() === 1` : (

I'm thinking what's missing is the information how many leases are currently in the process of being signed. If we added that then we can modify `canCreateMoreLeaseProcesses`
```ts
  private canCreateMoreLeaseProcesses() {
    return this.getSize() + this.leasesBeingSigned < this.maxPoolSize;
  }
```
The problem is that this variable needs to be incremented synchronously **before** we call `await this.createLeaseProcess()` because that would push it to the end of the microtasks queue and other `acquire`s would run before we increment the variable.
The version I have working looks like this:
```ts
  async acquire(): Promise<LeaseProcess> {
    if (this.isDraining) {
      throw new Error("The pool is in draining mode");
    }
    let leaseProcess = await this.takeValidLeaseProcess();
    if (!leaseProcess) {
      if (!this.canCreateMoreLeaseProcesses()) {
        return this.enqueueAcquire();
      }
      // new part here VVV
      this.leasesBeingSigned++;
      leaseProcess = await this.createNewLeaseProcess().finally(() => this.leasesBeingSigned--);
    }
    this.borrowed.add(leaseProcess);
    this.events.emit("acquired", leaseProcess.agreement);
    return leaseProcess;
  }
```
a bit ugly :( I'm open for other suggestions

mgordel:
@SewerynKras yes, you are right, this implementation solves the problem, I have tested it in several ways and it works fine.
I changed the code to your implementation

src/lease-process/lease-process-pool.test.ts
```diff
@@ -312,6 +335,7 @@ describe("LeaseProcessPool", () => {
       expect(pool["destroy"]).toHaveBeenCalledWith(lease3);
     });
     it("prevents new leases from being acquired during the drain", async () => {
+      jest.useFakeTimers();
```
SewerynKras, line 338: Isn't this a side-effect of enabling fake timers in that test above? IMO we should call `jest.useRealTImers` in `beforeEach` to "reset it"

SewerynKras (approved):

PR #977, opened 2024-06-13
feat: added ability to cancel lease processes

mgordel:
This PR includes:
 - removed the default timeout in draft-proposal-pool, instead added the `signalOrTimeout` parameter in the acquire method
 - added `signalOrTimeout` parameter to `oneOf` in golem-network
 - added `signalOrTimeout` parameter to `acquire` in lease-proposal-pool
 - changed `timeout` to `signalOrTimeout` parameter in `CommandOptions` in work-context
 - added `signalOrTimeout` parameter to `signAgreementFromPool` in market-module
 - removed `activityExecutionTimeout`  and `timeout` param in `execute` in exe-script-executor, instead added `signalOrTimeout` param in constructor (to cancel all executions) and in `execute` method (to cancel single execution)

### DIFF
--- a/src/activity/activity.module.ts
+++ b/src/activity/activity.module.ts
@@ -248,10 +248,7 @@ export class ActivityModuleImpl implements ActivityModule {
                 "Initializing of the exe-unit has been aborted due to a timeout",
                 options.signalOrTimeout.reason,
               )
-            : new GolemAbortError(
-                "Initializing of the exe-unit has been aborted by a signal",
-                options.signalOrTimeout.reason,
-              );
+            : new GolemAbortError("Initializing of the exe-unit has been aborted", options.signalOrTimeout.reason);
         throw error;
       }
       throw error;
--- a/src/activity/activity.module.ts
+++ b/src/activity/activity.module.ts
@@ -7,7 +7,7 @@ import { WorkContext, WorkOptions } from "./work";
 import { ExeScriptExecutor, ExeScriptRequest, ExecutionOptions } from "./exe-script-executor";
 import { Observable, catchError, tap } from "rxjs";
 import { StreamingBatchEvent } from "./results";
-import { GolemAbortError } from "../shared/error/golem-error";
+import { GolemAbortError, GolemTimeoutError } from "../shared/error/golem-error";
 
 export interface ActivityModule {
   events: EventEmitter<ActivityEvents>;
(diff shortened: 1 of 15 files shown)


tests/e2e/leaseProcessPool.spec.ts
```diff
@@ -174,4 +174,29 @@ describe("LeaseProcessPool", () => {
...
+  });
+
+  it("should finalize the lease process during execution", async () => {
+    const pool = glm.lease.createLeaseProcessPool(proposalPool, allocation, { replicas: 1 });
+    const leaseProcess = await pool.acquire();
+    const exe = await leaseProcess.getExeUnit();
+    return new Promise(async (res) => {
+      leaseProcess.events.on("finalized", async () => res(true));
+      setTimeout(() => leaseProcess.finalize(), 8_000);
+      await expect(exe.run("sleep 10 && echo Hello World")).rejects.toThrow(
```
SewerynKras, line 197: if the promise gets resolved before this rejects then jest won't catch the error, we should add `expect.assertions(1)` at the beginning of the test

src/activity/activity.module.ts
```diff
@@ -240,6 +241,19 @@ export class ActivityModuleImpl implements ActivityModule {
...
+            ? new GolemTimeoutError(
+                "Initializing of the exe-unit has been aborted due to a timeout",
+                options.signalOrTimeout.reason,
+              )
+            : new GolemAbortError(
+                "Initializing of the exe-unit has been aborted by a signal",
+                options.signalOrTimeout.reason,
+              );
+        throw error;
+      }
```
SewerynKras, line 256: Why is this here? `ctx.before` already throws `GolemAbortError`

SewerynKras (changes requested):

src/lease-process/lease-process.ts
```diff
@@ -51,26 +56,34 @@ export class LeaseProcess {
...
-
-    try {
-      this.logger.debug("Waiting for payment process of agreement to finish", { agreementId: this.agreement.id });
-      if (this.currentWorkContext) {
-        await this.activityModule.destroyActivity(this.currentWorkContext.activity);
-        if ((await this.fetchAgreementState()) !== "Terminated") {
-          await this.marketModule.terminateAgreement(this.agreement);
-        }
+    // Prevent this task from being performed more than once
+    return this.asyncLock.acquire("finalize", () => {
```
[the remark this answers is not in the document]
mgordel: yes, I am testing it again and the async lock is not needed. I added it when I was trying to fix starting finalization several times at the same time and left it unnecessarily

src/market/draft-offer-proposal-pool.ts
```diff
@@ -118,46 +105,45 @@ export class DraftOfferProposalPool {
   }
 
   /**
-   * Attempts to obtain a single proposal from the pool
-   *
-   * This method will reject if no suitable proposal will be found within {@link DraftOfferProposalPool.acquireTimeoutSec} seconds.
+   * Attempts to obtain a single proposal from the poolonds.
```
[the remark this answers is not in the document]
mgordel: 🙃

src/activity/activity.module.ts
mgordel: not all parts of `ctx.before` throw an `GolemAbortError` in this case. Some throw `GolemWorkError` with interrupted script execution in activity. It is true that in the `previous` field there will be AbortError from Activity, but I wanted to avoid inconsistency.

SewerynKras: ok, but in that case if someone passes a timeout (number) then this check won't run, it should be consistent. Have you considered checking the `ctx.abortSignal` directly?

mgordel: Yes, you are right. I moved the handling of this to `ctx.before` in the work-context and refactored it a bit. See now.

SewerynKras (approved):

PR #985, opened 2024-06-19
Rename lease to rental

SewerynKras:


mgordel (approved):
I didn't notice any changes in logic. Only name changes and comments. OK for me

PR #986, opened 2024-06-20
Example for setup and teardown

SewerynKras:


mgordel (approved):
Wow, really nice example. It starts with curiosity and surprise at the end :)

PR #989, opened 2024-06-20
JST-924: Rearrange docs in the repo for improved maintenance

grisha87:


README.md
```diff
@@ -163,280 +173,41 @@ const order: MarketOrderSpec = {
 })().catch(console.error);
```
SewerynKras, line 173: Looks like this example isn't updated (maybe some merge error), it should be `glm.oneOf({ order })` instead of `oneOf(order)`

README.md
```diff
@@ -163,280 +173,41 @@ const order: MarketOrderSpec = {
...
 
-It is recommended to run unit tests and static code analysis before committing changes.
-
-```bash
-yarn lint
-# and
-yarn format
-```
+In case you find an issue with the examples or the SDK itself, feel free to submit
+an [issue report](https://github.com/golemfactory/golem-js/issues) to the repository.
```
SewerynKras, line 211: Link to ./CONTRIBUTING.md needed

README.md
```diff
@@ -163,280 +173,41 @@ const order: MarketOrderSpec = {
...
 through them and learn about the recommended practices. All examples are automatically tested during our release
 process.
 
-In case you find an issue with the examples, feel free to submit
-an [issue report](https://github.com/golemfactory/golem-js/issues) to the repository.
-
 You can find even more examples and tutorials in
 the [JavaScript API section of the Golem Network Docs](https://docs.golem.network/docs/creators/javascript).
 
+## Documentation
```
SewerynKras, line 191: IMO link to the docs should be at the very top, even after installation

README.md
```diff
@@ -163,280 +173,41 @@ const order: MarketOrderSpec = {
...
-
-### Setup and teardown methods
+### Read more
 
-You can define a setup method that will be executed the first time a provider is rented and a teardown method
-that will be executed before the rental is done. This is useful when you want to avoid doing the same work
-multiple times when running multiple tasks on the same provider.
+- Read about [other available usage patterns](./docs/USAGE.md) to learn more on how you can leverage the SDK.
+- Check the [`examples` directory](./examples) for working examples showcasing various features of the SDK. All examples
+  are continuously E2E tested.
```
SewerynKras, line 180: This is basically a repeat of what's directly below, IMO we should just keep the ## Examples section and refator this into a ## Usage (or ## Features) section

```md
## Features
Read about [other available usage patterns](./docs/USAGE.md) to learn more on how you can leverage the SDK.

## Examples
...
```

README.md
```diff
@@ -14,32 +14,36 @@
...
+  - [Reporting issues](#reporting-issues)
   - [See also](#see-also)
+  <!-- TOC -->
+
+## Features
+
+Become a **Requestor** in the **Golem Network** and use this SDK to:
+
+- 🌐 Acquire compute resources from Providers using a convenient API
+- 🚢 Deploy run your workloads with these resources and get the results back to your machine
```
SewerynKras, line 40: Just "run your workloads", "deploy run" doesn't sound correct to me

SewerynKras (changes requested):

CONTRIBUTING.md
```diff
@@ -0,0 +1,9 @@
+# Contributing
```
SewerynKras, line 1: IMO a good contributing guide should have:
1. Local setup (clone repo + npm run dev to start rollup)
2. Running lint and tests
3. PR best practices (describe the feature, add unit tests if applicatble etc)

We can borrow this from someones repo, for example elysia has a nice short and to the point guide: https://github.com/elysiajs/elysia/blob/main/CONTRIBUTING.md

README.md
grisha87: Added

README.md
grisha87: I've decided to keep it here for now.

CONTRIBUTING.md
grisha87: I've used it as an inspiration. Provided something initial. We can evlove this over time.

README.md
```diff
@@ -163,280 +179,41 @@ const order: MarketOrderSpec = {
...
 
 ## Testing
 
-Read the dedicated [testing documentation](./TESTING.md) to learn more about how to run tests of the SDK.
+Read the dedicated [testing documentation](./docs/TESTING.md) to learn more about how to run tests of the SDK.
 
 ## Contributing
 
-It is recommended to run unit tests and static code analysis before committing changes.
+Read the [Contributing Guide](./CONTRIBUTING.md) for details on how you can get involved. In case you find an issue with the examples or the SDK itself, feel free to submit
```
SewerynKras, line 211: Why is `CONTRIBUTING.md` in root when all other markdown files are in `docs/`?

docs/UPGRADING.md
```diff
@@ -0,0 +1,254 @@
+# Upgrade Guide
+
+This document describes the breaking changes introduced in each major version of MyLibraryJS and the necessary steps you need to take to ensure your application continues to work correctly.
```
SewerynKras, line 3: `@golem-sdk/MyLibraryJS` when?

docs/UPGRADING.md
```diff
@@ -0,0 +1,254 @@
...
+```ts
+// after
+import { GolemNetwork } from "@golem-sdk/golem-js";
+
+(async function main() {
+  const glm = new GolemNetwork();
+  try {
+    await glm.connect();
+
+    const retnal = await glm.oneOf({
```
SewerynKras, line 122: TE by default does this work in parallel, but the provided snipped does it sequentially on one machine. We should promote `manyOf` for cases where multiple things need to be done at the same time, like you do later on

docs/CONCEPTS.md
```diff
@@ -0,0 +1,124 @@
...
+### Why is it needed
+
+When you obtain resources via the Golem Network, your _Golem Virtual Machine Image (GVMI for short)_ is going to be deployed by `golem-js` into the _Activity_ running on the Provider. Technically, on the Provider instantiates an _ExeUnit_ which is a physical implementation of the _Activity_ of the Golem Network Protocol. Without digging in too much details, it's this _ExeUnit_ which at the end performs the operations that you issue via the Golem Network.
+
+If you're familiar with Docker, you can picture the architecture in the following way:
+
+| Golem              | Docker       | Kubernetes   |
+| ------------------ | ------------ | ------------ |
+| GVMI               | Docker Image | Docker Image |
+| Activity (ExeUnit) | Container    | Container    |
```
SewerynKras, line 110: why even show k8s here if it's the same as Docker?

README.md
grisha87: Good hint, I thought that it's the pattern recommended by GH but after checking the best practices I see that this could be placed in the docs as well.

https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-readmes

docs/UPGRADING.md
grisha87: 🙈

docs/UPGRADING.md
grisha87: You're right, TE with `maxParallelTasks: 5 (the default)` will execute the tasks concurrently, while the new snippet will execute them in a sequential manner.

Nevertheless, the intent of this section was to show 1:1 transition between snippets specifically to show how the `Array.map` can be still used to execute a series of similar commands with different arguments. I'd like to keep the focus on this aspect.

Your observation is valueable and I will honour that by placing a note in the upgrade gude that compared to TE this is going to be executed sequentially, and that for parallel execution people should refer to `manyOf` example.

On a different note, it might be worth considering to fix the naming of the params for `manyOf` -> instead of `concurrency` I would simply use `count`.

Why? Concurrency and parallelism are two different terms. Technically we run parallel distributed "computations", so actually `parallelism` is the word we should use instead of `concurrency` (running on the same thread). 

I will create a ticket for rename of `oder.concurrency` to `order.count` for `manyOf`.

docs/CONCEPTS.md
grisha87: I'm dropping this comparision as it's not helpful at all and after talking with Witek from the Core Team I identified too many holes in this comparision.

docs/UPGRADING.md
```diff
@@ -84,6 +84,18 @@ Areas where the changes are needed:
...
+> `TaskExecutor` by default assumed that users want to run these commands in parallel, so it
+> was establishing up to 5 agreements with different providers and then sent out these
+> commands to different providers to leverage parallelism.
+>
+> With `golem-js@3.0`, when you use `oneOf` to rent resources, you are establishing one agreement
+> with a single provider and the commands which you issue with `Array.map` will be executed
+> sequentially by that single provider.
+>
+> If you want to achieve _parallel command execution_ on different providers, refer to the [Engaging with many providers at the same time](#engaging-with-many-providers-at-the-same-time) example.
+
```
SewerynKras, line 98: Why even show this in the migration if it's fundamentally different behavior? TE users only care about parallelism. Either drop the entire `Repeating commands with map of different parameter values` section or use `manyOf` here

SewerynKras (changes requested):

docs/UPGRADING.md
grisha87: Hmmmm. I guess you're right, this is more an example of the API usage than real migration. Thanks for that point of view :)

SewerynKras (approved):

PR #990, opened 2024-06-20
SignalOrTimeout param for ResourceRental

mgordel:


src/resource-rental/resource-rental.ts
```diff
@@ -74,7 +76,19 @@ export class ResourceRental {
...
           this.logger.info("Waiting for payment process of agreement to finish", { agreementId: this.agreement.id });
-          await waitForCondition(() => this.paymentProcess.isFinished());
+          const abortSignal = createAbortSignalFromTimeout(signalOrTimeout);
+          await waitForCondition(() => this.paymentProcess.isFinished(), {
+            signalOrTimeout: abortSignal,
+          }).catch((error) => {
+            this.paymentProcess.stop();
+            if (error instanceof GolemTimeoutError) {
+              throw new GolemTimeoutError(
+                `The finalization of payment process has been aborted due to a timeout (${signalOrTimeout}ms)`,
```
SewerynKras, line 86: isn't `${signalOrTimeout}ms` going to end up as `[object AbortSignal]ms`?

src/resource-rental/resource-rental.ts
```diff
@@ -93,21 +107,35 @@ export class ResourceRental {
...
+    this.abortController.signal.addEventListener("abort", () =>
+      abortController.abort(this.abortController.signal.reason),
+    );
+    if (signalOrTimeout) {
+      const abortSignal = createAbortSignalFromTimeout(signalOrTimeout);
+      abortSignal.addEventListener("abort", () => abortController.abort(abortSignal.reason));
+      if (signalOrTimeout instanceof AbortSignal && signalOrTimeout.aborted) {
+        abortController.abort(signalOrTimeout.reason);
+      }
+    }
```
SewerynKras, line 131: why do we need a new AbortController instance here? Can't we just pass `abortSignal` to `createExeUnit`?

src/shared/utils/wait.ts
```diff
@@ -1,23 +1,28 @@
...
  * @return {Promise<void>} - Resolves when the condition is met or rejects with a timeout error if it wasn't met on time.
  */
 export function waitForCondition(
   check: () => boolean | Promise<boolean>,
-  opts?: { timeoutSeconds?: number; intervalSeconds?: number },
+  opts?: { signalOrTimeout?: number | AbortSignal; intervalSeconds?: number },
 ): Promise<void> {
-  const timeoutSeconds = opts?.timeoutSeconds ?? 30;
+  const DEFAULT_TIMEOUT_SECONDS = 30;
+  const abortSignal =
```
SewerynKras, line 18: can't we use `createAbortSignalFromTimeout` here instead of these nested ternaries?
```ts
  const signal =
    opts?.signalOrTimeout !== undefined
      ? createAbortSignalFromTimeout(opts?.signalOrTimeout)
      : AbortSignal.timeout(DEFAULT_TIMEOUT_SECONDS * 1000);
```

SewerynKras (changes requested):

src/resource-rental/resource-rental.ts
mgordel: Actually, there is probably only one situation in which the type can be `[object AbortSignal]`, when the user provides `AbortSignal.timeout()` as a parameter. In all other cases, `GolemTimeoutError` thrown by `waitForCondition` will only occur if the user specifies `signalOrTimeout` as a number. But I guess because of this one exception we have to remove this information about the timeout value.

src/resource-rental/resource-rental.ts
mgordel: Because `getExeUnit()` should be aborted in two cases, via `signalOrTimeout` from the user parameter or via `abortController` from `ResourceRental` (e.g. if someone calls `stopAndFinalize()`.
That's why we need to listen to these two sources.

SewerynKras: Ok that makes sense.
Shame that `AbortSignal.any([signal1, signal2, ...])` is only from node 20 and up, it would be nice here

src/shared/utils/wait.ts
mgordel: I don't think so, because `opts?.signalOrTimeout` is given in sec instead of ms and we still have to add checking if there is a number and multiply by 1000, so I don't know what is better...?

src/resource-rental/resource-rental.ts
mgordel: Yes `AbortSignal.any` would be cool in this situation

src/shared/utils/wait.ts
SewerynKras: `signalOrTimeout` should be in milliseconds then, let's be consistent across the codebase with this

mgordel: ok, we agree 🙂

SewerynKras (approved):

PR #991, opened 2024-06-21
feat: make `glm.disconnect` `rentalPool.drainAndClear` and `rental.stopAndFinalize` idempotent

SewerynKras:
- make `glm.disconnect` idempotent
- make `pool.drainAndClear` idempotent
- make `rental.stopAndFinalize` idempotent
- interrupt `oneOf` and `manyOf` if `glm` gets disconnected while it's being created
- interrupt rentals being created in the pool when it gets drained


src/golem-network/golem-network.test.ts
```diff
@@ -187,4 +187,19 @@ describe("Golem Network", () => {
...
+    it("reuses the same promise if called multiple times", async () => {
+      const glm = getGolemNetwork();
+      const glmSpy = spy(glm);
+      when(glmSpy["startDisconnect"]()).thenResolve();
+      expect(glm["disconnectPromise"]).toBeUndefined();
+      const promise1 = glm.disconnect();
+      const promise2 = glm.disconnect();
+      const promise3 = glm.disconnect();
+      expect(glm["disconnectPromise"]).toBeDefined();
+      await Promise.all([promise1, promise2, promise3]);
```
mgordel, line 200: Parallel runs passes the test, but sequential fails:
```ts
await promise1;
await promise2;
```

src/golem-network/golem-network.ts
```diff
@@ -314,20 +315,35 @@ export class GolemNetwork {
...
-
-    this.services.proposalCache.flushAll();
-
-    this.events.emit("disconnected");
-    this.hasConnection = false;
+    if (this.disconnectPromise) {
+      return this.disconnectPromise;
+    }
+    this.disconnectPromise = this.startDisconnect().finally(() => {
+      this.disconnectPromise = undefined;
```
mgordel, line 344: why is this needed? this means that running disconnect sequentially will not be idempotent
```ts
await glm.disconnect();
await glm.disconnect(); // runs disconnect again causes another attempt to releas allocation etc.
```
so maybe it should be `undefined` only when the user run the `connect` again?

src/resource-rental/resource-rental-pool.ts
```diff
@@ -266,20 +289,14 @@ export class ResourceRentalPool {
...
-    this.highPriority.clear();
-    this.borrowed.clear();
-    this.isDraining = false;
-    this.events.emit("end");
-    return;
+    if (this.isDraining) {
+      return this.drainPromise;
+    }
+    this.drainPromise = this.startDrain().finally(() => {
+      this.drainPromise = undefined;
```
mgordel, line 296: same behavior as above

src/golem-network/golem-network.test.ts
SewerynKras: This is the correct logic in my opinion, if you await the promise then the disconnect finishes and on the next run it's the same as running `disconnect` on a newly created `GolemNetwork` (without `connect` first), nothing should happen in that case

src/golem-network/golem-network.ts
SewerynKras: `cleanupTasks` should be set to `null` after disconnect finishes, this is a bug

src/golem-network/golem-network.test.ts
SewerynKras: What I mean is I think this should be valid:
```ts
const glm = new GolemNetwork()
await glm.connect();
await glm.disconnect();
await glm.disconnect() // nothing happens
await glm.connect();
await glm.disconnect();
```

mgordel: ok, I agree, so disconnecting on an unconnected `GolemNetwork` should not result in any action on the network, so it should either be resolved immediately or give an error, e.g. `GolemUserError("Network is not connected")`

SewerynKras: same for rental pool
```ts
const pool = new ResourceRentalPool({ ... });
await pool.ready();
await pool.drainAndClear(); // finalizes all rentals
await pool.drainAndClear(); // nothing happens
await pool.ready();
// pool is ready to use again
await pool.drainAndClear(); // finalizes all rentals
```

mgordel (approved):

PR #1002, opened 2024-06-27
Replace readable with observables

SewerynKras:
Breaking changes:
- `executor.execute` is no longer an async function that sends the script and creates a Readable. Instead, it synchronously creates an observable that once subscribed to, sends the script and starts emitting it's results
- `RemoteProcess` (result of `runAndStream`) now uses `Subject`s instead of `Readable`s to implement `stdout` and `stderr`
- Removed stream polyfill as it's no longer used


src/activity/exe-unit/batch.ts
```diff
@@ -136,10 +135,10 @@ export class Batch {
...
   }
 
-  async endStream(): Promise<Readable> {
+  async endStream(): Promise<Observable<Result>> {
     const script = this.script;
     await script.before();
-    let results: Readable;
+    let results: Observable<Result>;
     try {
       results = await this.executor.execute(this.script.getExeScriptRequest());
```
grisha87, line 143: Redundant await. At the same time that's no longer throwing as it's returing an Observable to she try/catch block here is also redundant.

src/activity/exe-unit/exe-unit.ts
```diff
@@ -392,16 +390,15 @@ export class ExeUnit {
     await sleep(100, true);
 
     // Send script.
-    const results = await this.executor.execute(
+    const results = this.executor.execute(
```
grisha87, line 393: Style: check the places where you run `executor.execute` and make sure you use the `$` (like `results$`) for consistency and readability.

src/shared/utils/rxjs.ts
```diff
@@ -0,0 +1,23 @@
...
+ * while `merge` would only complete when _all_ observables complete.
+ */
+export function mergeUntilFirstComplete<FirstObsType, SecondObsType>(
+  observable1: Observable<FirstObsType>,
+  observable2: Observable<SecondObsType>,
+): Observable<FirstObsType | SecondObsType> {
+  const completionSubject = new Subject<void>();
+  const observable1WithCompletion = observable1.pipe(
+    takeUntil(completionSubject),
+    finalize(() => completionSubject.next()),
```
grisha87, line 15: Style: `completionSubject.complete()` would have the same effect, but would be easiert to understand at first glance, I had to spend a minute or two to understand what's going on here.

src/activity/exe-script-executor.ts
```diff
@@ -63,186 +68,169 @@ export class ExeScriptExecutor {
       abortSignal.addEventListener("abort", () => abortController.abort(abortSignal.reason));
     }
 
-    try {
-      abortController.signal.throwIfAborted();
-      batchId = await this.send(script);
-      batchSize = JSON.parse(script.text).length;
+    // observable that emits when the script execution should be aborted
+    const abort$ = new Observable<never>((subscriber) => {
```
grisha87, line 72: Logic: Do we really need this in form of a observable? What I'm thinking about instead of this is that we keep only the `sendScript$` observable, which respects the abort singal and once the signal is raised the `sencScript$` would `.error` and `.complete` after that.

This would remove the `abort$` and the need for `mergeUntilFirstComplete` altogether.

src/activity/exe-script-executor.ts
```diff
@@ -63,186 +68,169 @@ export class ExeScriptExecutor {
...
-        `Unable to execute script. ${message}`,
-        WorkErrorCode.ScriptExecutionFailed,
-        this.activity.agreement,
-        this.activity,
-        this.activity.provider,
-        error,
-      );
-    }
+    // observable that when subscribed to, will execute the provided script and emit results
+    const sendAndStreamResults$ = sendScript$.pipe(mergeMap(getResults));
```
grisha87, line 126: I think this is a bit over-engineered. sendScript$ is an observable that will produce a single value which is the `{batchId: string; batchSize: number}` and it will be used to `getResults`.

Observables are designed to _emit multiple values over time_. If `sendScript$` ever only emits one, then it should have be a simple promise. Then the `streamingBatch` or `pollingBatch` observables can be invoked to deliver the observable.

I do see that would change the return type of `execute` to `Promise<Observable<Result>>`, but I'm actually more fine with that compared to the _exploit_ done with `sendScript$`. That's _smart_, but I would prefer _simple_ here :)

grisha87 (approved):

grisha87 (changes requested):
As per remarks in the comments :)

src/shared/utils/rxjs.ts
SewerynKras: `completionSubject.complete` doesn't trigger `takeUntil` to stop, only `next`.

src/activity/exe-script-executor.ts
SewerynKras: This is still needed if we want the resulting observable to emit an error when timeout is reached

grisha87 (approved):

PR #1003, opened 2024-06-28
fix(market): fixed demand refresh logic that produced unusable counte…

grisha87:
…r-offers and agreements

## Changed

* added MarketModuleOptions, and allowed configuring demandRefreshIntervalSec
* switched publishAndRefresh demand to use this setting instead of the expiry time of the demand (deprecated and can be removed)
* the `golem.srv.comp.expiration` property of the agreement offer is now set by the workload director and not the basic director
* the allocation is now created with +15min more expiry time than the market order itself

## Fixed

* fixed the issue where we fail to accept the invoice for an agreement that was terminated due to expiry with an allocation that expired together with the agreement (change: the allocation expiry time is set to +15 minutes more than the configured rental duration)

### DIFF
--- a/src/market/market.module.ts
+++ b/src/market/market.module.ts
@@ -26,7 +26,7 @@ import {
   OfferProposalFilter,
   ProposalsBatch,
 } from "./proposal";
-import { BuildDemandOptions, DemandBodyBuilder, DemandSpecification } from "./demand";
+import { OrderDemandOptions, DemandBodyBuilder, DemandSpecification } from "./demand";
 import { IActivityApi, IFileServer } from "../activity";
 import { StorageProvider } from "../shared/storage";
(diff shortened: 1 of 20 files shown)


src/market/market.module.ts
```diff
@@ -494,7 +521,7 @@ export class MarketModuleImpl implements MarketModule {
         });
         return agreement;
       } catch (error) {
-        this.logger.debug("Failed to propose agreement, retrying", { error });
+        this.logger.warn("Failed to propose agreement, retrying", { error });
```
SewerynKras, line 524: this can happen often and is not an error, it's totally ok for a provider to reject our offer if they found someone else in the meantime, that's why this method was created in the first place. No need to spam the user with warnings that are not actionable and not something they should be worried about. `debug` level is correct here

SewerynKras (changes requested):

SewerynKras (approved):

PR #1005, opened 2024-06-28
feat: added missing events in the resource rental module

mgordel:


src/resource-rental/resource-rental.ts
```diff
@@ -11,10 +11,20 @@ import { MarketModule } from "../market";
...
+  exeUnitCreated: (activity: Activity) => void;
+
+  /** Emitted when there is an error while creating the ExeUnit */
+  errorCreatingExeUnit: (error: Error) => void;
+
+  /** Emitted when the ExeUnit is successfully destroyed */
+  exeUnitDestroyed: (activity: Activity) => void;
+
+  /** Emitted when there is an error while destroying the ExeUnit */
+  errorDestroyingExeUnit: (error: Error) => void;
```
grisha87, line 27: Would just keep this as `error` and then leave the details of what happened to the error itself.

src/resource-rental/resource-rental.ts
```diff
@@ -11,10 +11,20 @@ import { MarketModule } from "../market";
...
-   * Raised when the rental process is fully finalized
-   */
+  /** Emitted when the rental process is fully finalized */
   finalized: () => void;
+
+  /** Emitted when ExeUnit is successfully created and initialised */
+  exeUnitCreated: (activity: Activity) => void;
+
+  /** Emitted when there is an error while creating the ExeUnit */
+  errorCreatingExeUnit: (error: Error) => void;
```
grisha87, line 21: Would just keep this as `error` and then leave the details of what happened to the error itself.

grisha87 (changes requested):

grisha87 (approved):

Issue #127, opened 2021-02-12
No zk payment scheduled after C-c on mainnet


ederenn:
mf | macOS BigSur | yajsapi 0.3.0-alpha.7 | No zk payment scheduled after C-c

[yajsapi-blender-term.log](https://github.com/golemfactory/yajsapi/files/5971905/yajsapi-blender-term.log)
[yagna_rCURRENT (3).log](https://github.com/golemfactory/yajsapi/files/5971902/yagna_rCURRENT.3.log)

same on js Windows
ederenn, Win 10 yajsapi 0.3.0-alpha.7

[yagna_rCURRENT.log](https://github.com/golemfactory/yajsapi/files/5971910/yagna_rCURRENT.log)
[yajsapi-current.log](https://github.com/golemfactory/yajsapi/files/5971911/yajsapi-current.log)
[yajsapi.log](https://github.com/golemfactory/yajsapi/files/5971912/yajsapi.log)

filipgolem:
Commit fdcad356 should fix this issue.

maaktweluit:
please note windows kills subtree of processes differently then posix..

could be this can never be fully fixed

filipgolem:
> please note windows kills subtree of processes differently then posix..
> 
> could be this can never be fully fixed

Payments after ctrl+c mentioned in this issue are fixed (commit fdcad35). Ctrl+c causes cancellation token to be cancelled - which stops everything gracefully in `yajsapi` - no processes are abruptly ended, this is not OS-dependent.

The only problem is that `node` returns to the terminal (i.e. you can issue commands, at least on macOS) when Ctrl+C is pressed - even though the signal handler is present and everything is shutting down correctly.

mateuszsrebrny:
Ok, so I undertand won't fix :)

PR #1021, opened 2024-07-09
fix(job): fixed rental finalization after job completed

mgordel:


src/experimental/job/job.ts
```diff
@@ -151,7 +151,7 @@ export class Job<Output = unknown> {
 
     signal.addEventListener("abort", onAbort, { once: true });
 
-    return workOnGolem(exeUnit);
+    return workOnGolem(exeUnit).finally(() => rental.stopAndFinalize());
```
SewerynKras, line 154: IMO this should be moved to the `finally` block on 127 and we should only emit the `ended` event once the rental is actually finalized

mgordel: there is no information about the `rental` in the `finally` block (#127), it has not been created yet. I added `try / finally` block and `await` before `finalize` so that event`ended` would be emitted always after rental is finalized

SewerynKras: ah you're right, my mistake. I like the change you did. I think we don't event need to call `waitAndFinalize` in `onAbort` because it will be called anyways after the error is thrown

src/experimental/job/job.ts
```diff
@@ -139,17 +139,12 @@ export class Job<Output = unknown> {
...
-      };
-
       if (signal.aborted) {
-        await onAbort();
+        this.events.emit("canceled");
         throw new GolemAbortError("Canceled");
       }
 
-      signal.addEventListener("abort", onAbort, { once: true });
+      signal.addEventListener("abort", () => this.events.emit("canceled"), { once: true });
```
SewerynKras, line 147: The issue now is that when the work is already started, aborting the job will **not** stop it. The unit test fails to catch this because the signal gets aborted before the actual work begins, you can see this if you modify the test to use a `jest fn`:
```ts
const taskFn = jest.fn().mockImplementation(() => {
  return new Promise((resolve) => setTimeout(resolve, 10000));
});
job.startWork(taskFn);

// flush promises
// await new Promise(setImmediate); // <------------- uncomment to cause a timeout
await job.cancel();
await expect(job.waitForResult()).rejects.toThrow("Canceled");

verify(mockRental.stopAndFinalize()).once();
expect(taskFn).toHaveBeenCalled(); // <---------------- error! 0 calls
```
(the abort error is thrown on line 144)
If you uncomment the `Promise(setImmediate)` then the test will time out, because the task function will start and aborting it won't do anything.

To fix this, I think we should race the `workOnGolem` promise with the abort signal, something like
```ts
const workPromise = workOnGolem(exeUnit);
const abortPromise = new Promise<never>((resolve, reject) => {
  signal.addEventListener(
    "abort",
    () => {
      this.events.emit("canceled");
      reject(new GolemAbortError("Canceled"));
    },
    { once: true },
  );
});
return Promise.race([workPromise, abortPromise]);
```

SewerynKras (changes requested):

src/experimental/job/job.ts
mgordel: I'm not sure if this is needed. All we need to do is pass the `signal` to `getExeUnit` method and then it will be passed to the activity executor and each attempt to do something or fetch the result will end with a `GolemAbortError`, which will stop the work.

SewerynKras: You're right, I didn't think of that

SewerynKras (approved):

PR #1028, opened 2024-07-11
fix(resource-rental): fixed error handling when creating an exe-unit

mgordel:
This unhandled promise caused unhandled rejection if the signal was aborted before `getExeUnit()` was executed. I added `null` for the catch because another catch will handle logging and triggering the error event. However, error propagation should only take place if this method is called via `getExeUnit()`. Automatic creation of exe-unit from the constructor should ignore the error.

### DIFF
--- a/src/resource-rental/resource-rental.ts
+++ b/src/resource-rental/resource-rental.ts
@@ -53,7 +53,9 @@ export class ResourceRental {
   ) {
     this.networkNode = this.resourceRentalOptions?.networkNode;
 
-    this.createExeUnit(this.abortController.signal).catch(() => null);
+    this.createExeUnit(this.abortController.signal).catch((error) =>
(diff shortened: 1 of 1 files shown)


src/resource-rental/resource-rental.ts
```diff
@@ -53,7 +53,7 @@ export class ResourceRental {
   ) {
     this.networkNode = this.resourceRentalOptions?.networkNode;
 
-    this.createExeUnit(this.abortController.signal);
```
grisha87, line 56: I've read the explanation which you provided on the PR. Still, I'd recommend gracefully logging the error on `debug` level for the sake of troubleshooting in future.

grisha87 (changes requested):

grisha87 (approved):

PR #1039, opened 2024-07-31
Feature/golem stats improvements

grisha87:
This PR implements improvements identified when building Golem Stats 2 website.


src/market/scan/scanned-offer.ts
```diff
@@ -51,34 +56,104 @@ export class ScannedOffer {
...
+   * }
+   * ```
+   */
+  get paymentPlatformAddresses(): PaymentPlatformAddressSet {
+    const platformProps = Object.entries(this.model.properties).filter(([key]) =>
+      key.startsWith("golem.com.payment.platform."),
+    );
+
+    const platformAddress = platformProps.map(([key, address]) => [
+      key.replace("golem.com.payment.platform.", "").replace(".address", ""),
```
SewerynKras, line 153: if someone sets a constraint that starts with `golem.com.payment.platform` but ends with something different from `.<platform>.address` then this method will produce nonsense. We should filter the properties not only with `startsWith` but also with `endsWith` or even better, with a single regex like
`/^golem\.com\.payment\.platform\.[a-zA-Z\-0-9]+\.address$/` (untested)

SewerynKras (changes requested):

src/market/scan/scanned-offer.ts
```diff
@@ -145,14 +145,15 @@ export class ScannedOffer {
...
-    const platformProps = Object.entries(this.model.properties).filter(([key]) =>
-      key.startsWith("golem.com.payment.platform."),
-    );
-
-    const platformAddress = platformProps.map(([key, address]) => [
-      key.replace("golem.com.payment.platform.", "").replace(".address", ""),
-      address,
-    ]);
+    const propRegex = /golem.com.payment.platform.([a-z0-9-]+).address/;
+    const platformProps = Object.entries(this.model.properties).filter(([key]) => key.match(propRegex));
```
SewerynKras, line 149: Is this first filter needed anymore? You're filtering again based on the same condition below

src/market/scan/scanned-offer.ts
```diff
@@ -145,14 +145,15 @@ export class ScannedOffer {
...
   get paymentPlatformAddresses(): PaymentPlatformAddressSet {
-    const platformProps = Object.entries(this.model.properties).filter(([key]) =>
-      key.startsWith("golem.com.payment.platform."),
-    );
-
-    const platformAddress = platformProps.map(([key, address]) => [
-      key.replace("golem.com.payment.platform.", "").replace(".address", ""),
-      address,
-    ]);
+    const propRegex = /golem.com.payment.platform.([a-z0-9-]+).address/;
```
SewerynKras, line 148: 1. `.` should be escaped, right now it matches any character (try `golem☠com☠payment☠platform☠nonsense☠address`)
2. are we sure all constraints keys are lowercase only?

grisha87: AD 1. Will fix that.
AD 2. Yes, that's the case.

src/market/scan/scanned-offer.ts
grisha87: As per our call: It's not needed, I will drop it.

SewerynKras (approved):

PR #1041, opened 2024-08-01
Fix TcpProxy for non HTTP communication

grisha87:
## The problem

The current implementation of the TCP proxy failed to correctly transfer data back and forth for non HTTP communication. Trying to execute plain TCP use cases with this rendered it unusable and errornous.

## Solution

This PR solves this by reimplementing the TCP Proxy logic and providing missing features such as buffering of the communication while waiting for the other endpoints to be fully ready to accept communication.

In addition:

- I've added a vanila golem-js example for the TcpProxy feature which was missing
- I've renamed the `waitForCondition` internal helper function to `waitFor` and exposed this as a utility from `golem-js` as this is a feature that gets implemented by many requestor scripts on their own.


examples/rental-model/advanced/tcp-proxy/tcp-proxy.ts
```diff
@@ -0,0 +1,86 @@
...
+    };
+
+    process.on("SIGINT", () => {
+      stopServer()
+        .then(() => rental.stopAndFinalize())
+        .then(() => logger.info("Shutdown routine completed"))
+        .catch((err) => logger.error("Failed to shutdown cleanly", err));
+    });
+
+    await waitFor(() => server.isFinished());
```
mgordel, line 80: When I test this example, calling SIGINT does not make `server.isFinished()` `true` and I always get a reject on `wait`, even though the rental is finalized. The `isFinished()` method is true only when the command called in `runAndStrem` finishes, which does not happen.

So we have to wait for something else or add a mechanism to interrupt runAndStream() - the stop method, which could only delete the activity, because there is no `delete batch`.

mgordel (changes requested):
I tested it via `telnet` and it works fine. But the example has a small bug

You can also add this example to package.json in examples.

examples/rental-model/advanced/tcp-proxy/tcp-proxy.ts
grisha87: I've adjusted the logic of `isFinished` to correctly track if the process is still "considered as runnning/finished". Also, I've simplified the waitFor implementation to not assume any timeout of 30s for the user which has been our sin in TE and we adopted the policy to not impose such assumptions on our users.

PR #1052, opened 2024-08-26
fix(resource-rental): fixed multiple triggering of finalization (including teardown)

mgordel:


grisha87:
As discussed during our call:

- Adjust the implementation of `stopAndFinalize` to not remove the `finalizePromise`
- Adjust the comment to the method, to state ``If the rental is already finalized, it will resolve immediately with the last finalization result.`

grisha87 (approved):

PR #1062, opened 2024-09-02
fix(demand): defined minimum expiration value for demand

mgordel:


grisha87 (changes requested):
As discussed, please issue a warning when the value will be too smal, but accept the user's will.

grisha87 (approved):

grisha87 (approved):

PR #1071, opened 2024-09-04
docs(deposit): added example demonstrating the use of allocation with deposit

mgordel:


package.json
```diff
@@ -133,7 +133,8 @@
     "typedoc-plugin-markdown": "^3.17.1",
     "typedoc-plugin-merge-modules": "^5.1.0",
     "typedoc-theme-hierarchy": "4.1.2",
-    "typescript": "^5.3.3"
+    "typescript": "^5.3.3",
+    "viem": "^2.21.1"
```
grisha87, line 137: I think this should be added only to the examples?

grisha87 (approved):
Everything looks good. I just came to the conclusion that there's a lot that the users have to do in order to really use the feature - for the purpose of the example you had to create few functions wrapping the complexity of the process. These are candidates to be placed in the SDK itself which we should explore as a potential feature set for the SDK.

mgordel:
definitely worth considering. I see it as `golem sdk deposit create --amount --fee --expiration` ...

PR #1073, opened 2024-09-05
feat(golem-network): added getIdentity method to GolemNetwork object

mgordel:


src/golem-network/golem-network.ts
```diff
@@ -188,6 +189,12 @@ export type GolemServices = {
   storageProvider: StorageProvider;
 };
 
+export type Identity = {
```
grisha87, line 192: We shouldn't blindly forward the data without any documentation this way. It's not clear for the user of this API what glm.getIdentity().identity means as it's `identity->identity` and there's not tsdoc that explains particular fields to the user.

I've tested the code and get the following result from the `getIdentity()`:

```ts
{
  identity: '0x46b1...',
  name: 'eth-warsaw-workshop-test',
  role: 'manager'
}
```

This is roughly similar to output of `yagna app-key show eth-warsaw-workshop-test --json`:

```ts
{
  "allowOrigins": [],
  "createdDate": "2024-08-30T08:29:07.361270875",
  "identity": "0x46b1...",
  "key": "ec...e3",
  "name": "eth-warsaw-workshop-test",
  "role": "manager"
}
```

I guess that the naming of things is broken on yagna's API and in fact we have a mixture of `identity` and `app-key` concepts, where an app-key belongs to an  `identity`.

As far as I'm aware, "identity" of the "Golem User" is established by his ETH address. This is reflected by `yagna id show`:

```ts
{
  "Ok": {
    "alias": null,
    "deleted": false,
    "isDefault": true,
    "isLocked": false,
    "nodeId": "0x46b1..."
  }
}
```

So actually, I'd rather expect from `glm.getIdentity(): string -> "0x46b1..."` instead of the object which melts together the identity and app-key information.

grisha87 (changes requested):

src/golem-network/golem-network.ts
mgordel: Yes, it's a bit confusing. I wanted to be consistent with yagna api. And the entity should probably contain fields: `{nodeId, name, role}` -> https://github.com/golemfactory/ya-client/blob/master/model/src/identity.rs

but for our purposes I agree that a single nodeId string might be better

PR #1075, opened 2024-09-06
docs(examples): added whitelist outbound example

grisha87:


examples/rental-model/advanced/outbound/whitelist/read-golem-js-releases.ts
```diff
@@ -0,0 +1,84 @@
...
+          },
+        },
+      },
+      signalOrTimeout: timeoutSignal,
+    });
+
+    timeoutSignal.removeEventListener("abort", onTimeout);
+
+    const exe = await rental.getExeUnit();
+    console.log(await exe.run("curl https://registry.npmjs.org/-/package/@golem-sdk/golem-js/dist-tags"));
```
mgordel, line 76: maybe it's better to show only stdout instead of the whole result object. curl displays strange info in stderr during connection

mgordel (approved):
We could add this example to our tests, but it would require copying the whitelist configuration for providers in Docker just like it is in TaskExecutor:
https://github.com/golemfactory/golem-sdk-task-executor/blob/master/tests/docker/setProviderWhitelist.sh

